Issue #2, opened 2021-10-29
Generation fetcher component


piodul:
Scylla doc page about generations: https://docs.scylladb.com/using-scylla/cdc/cdc-stream-generations/

We need a component which will provide us with information about generations. This information will be important for the library to know from which streams it should be reading at the moment, learn about new generations and switch to reading the streams from the new generation.

It should provide us with the following information/support the following operations:

- What are the currently known generations,
- For a given timestamp, which generation contains it, i.e. if there was a row in CDC log with that timestamp - which generation would contain it,
- For a given generation, what is the the next generation,
- For a given generation timestamp, fetch stream IDs of that generation.

Currently, we are interested only in the newest format of generation tables, introduced in Scylla 4.4 - so we don't need to support the legacy `cdc_streams_descriptions` table.

Issue #4, opened 2021-10-29
Implement a simple CDC log printer


piodul:
Depends on #2 and #3

The task is to implement a simple version of the library which reads data from a specified CDC log table and prints it to the standard output. The library should start reading starting from a point of time specified by the user and should correctly handle generation changes. The library should process data from all streams.

Note: although in #3 it is mentioned that the stream reader should be able to read from multiple streams at once, I recommend creating a separate stream reader for each stream for now. It is beneficial to read from multiple streams which belong to a single vnode, but it can be done as a later optimization.

Ponewor:
What is the expected behaviour of the library when there is no generation for a given point of time?

piodul:
It should start consuming from the beginning of the generation which has the lowest timestamp. If there are no generations at all, it should wait in a loop, periodically checking for the first generation to appear.

You can use generation component's method which fetches a list of all generations and then choose the lowest one. I know that it isn't particularly efficient, but let's use the interface defined in #2 for now. We can optimize it later, after this issue is closed.

PR #6, opened 2021-11-04
Generation Fetcher

macher259:
My take on [issue](https://github.com/piodul/scylla-cdc-rust/issues/2).
I have added some tests thanks to your idea from the slack, but I am not sure if I got what you meant right. 

Probably it's still WIP, but I would like to hear some feedback.

Tests compare output of the functions with output of the queries made in test tables that were populated with the data from my local database. 

Feel free to roast me.
Fixes: #2

### DIFF
--- a/scylla-cdc/src/stream_generations.rs
+++ b/scylla-cdc/src/stream_generations.rs
@@ -1,0 +1,531 @@
+use futures::stream::StreamExt;
+use scylla::batch::Consistency;
+use scylla::frame::response::result::Row;
+use scylla::frame::value::Timestamp;
+use scylla::query::Query;
+use scylla::{IntoTypedRows, Session};
+use std::error::Error;
+use std::sync::Arc;
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Generation(chrono::Duration);
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct StreamID(Vec<u8>);
+
+// Component responsible for managing stream generations.
+pub struct GenerationFetcher {
+    generations_table_name: String,
+    streams_table_name: String,
+    session: Arc<Session>,
+}
+
+// Number taken from: https://www.scylladb.com/2017/11/17/7-rules-planning-queries-maximum-performance/.
+const DEFAULT_PAGE_SIZE: i32 = 5000;
+
+impl GenerationFetcher {
+    // Creates new instance of GenerationFetcher.
+    pub fn new(session: &Arc<Session>) -> GenerationFetcher {
+        GenerationFetcher {
+            generations_table_name: "system_distributed.cdc_generation_timestamps".to_string(),
+            streams_table_name: "system_distributed.cdc_streams_descriptions_v2".to_string(),
+            session: Arc::clone(session),
+        }
+    }
+
+    // Function instead of constant for testing purposes.
+    fn get_all_stream_generations_query(&self) -> String {
+        format!(
+            r#"
+    SELECT time
+    FROM {}
+    WHERE key = 'timestamps';
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    pub async fn fetch_all_generations(&self) -> Result<Vec<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_all_stream_generations_query());
+        self.select_consistency(&mut query).await?;
+        query.set_page_size(DEFAULT_PAGE_SIZE);
+
+        let mut rows = self
+            .session
+            .query_iter(query, &[])
+            .await?
+            .into_typed::<(chrono::Duration,)>();
+
+        let mut generations = vec![];
+
+        while let Some(next_row) = rows.next().await {
+            let (generation,) = next_row?;
+            generations.push(Generation(generation));
+        }
+
+        Ok(generations)
+    }
+
+    // Function instead of constant for testing purposes.
+    fn get_generation_by_timestamp_query(&self) -> String {
+        format!(
+            r#"
+    SELECT time
+    FROM {}
+    WHERE key = 'timestamps'
+    AND time <= ?
+    LIMIT 1;
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    // Given a timestamp of an operation, fetch generation that contains this operation.
+    pub async fn fetch_generation_by_timestamp(
+        &self,
+        time: &chrono::Duration,
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_generation_by_timestamp_query());
+        self.select_consistency(&mut query).await?;
+
+        let result = self.session.query(query, (Timestamp(*time),)).await?.rows;
+
+        GenerationFetcher::return_single_row(result)
+    }
+
+    // Function instead of constant for testing purposes.
+    fn get_generation_by_timeuuid_query(&self) -> String {
+        format!(
+            r#"
+    SELECT time
+    FROM {}
+    WHERE key = 'timestamps'
+    AND time <= dateOf( ? )
+    LIMIT 1;
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    // Given a timeuuid of an operation, fetch generation that contains this operation.
+    pub async fn fetch_generation_by_timeuuid(
+        &self,
+        time: &Uuid,
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_generation_by_timeuuid_query());
+        self.select_consistency(&mut query).await?;
+
+        let result = self.session.query(query, (time,)).await?.rows;
+
+        GenerationFetcher::return_single_row(result)
+    }
+
+    // Function instead of constant for testing purposes.
+    fn get_next_generation_query(&self) -> String {
+        format!(
+            r#"
+    SELECT time
+    FROM {}
+    WHERE key = 'timestamps'
+    AND time > ?
+    ORDER BY time DESC
+    LIMIT 1;
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    pub async fn fetch_next_generation(
+        &self,
+        generation: &Generation,
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_next_generation_query());
+        self.select_consistency(&mut query).await?;
+
+        let result = self
+            .session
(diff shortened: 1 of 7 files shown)


piodul (changes requested):
Some other issues to address:

- I think it would be nice to introduce a type which wraps all the necessary context for querying generation/stream info, and change the functions introduced here into methods of that type. You can capture the session object as `Arc<Session>`, and keep table names as `String`s.
- Try reducing repetition of magic numbers/strings in the test code. Ideally, stream IDs/timestamps/etc. should be given clear names and declared as constants at the beginning of the `tests` module.
- Before submitting a PR please make sure that it is cleanly split into commits. Each commit should be a separate, logical step, so that the reviewing process is easier. In this case, I think squashing everything into a single commit should be the right thing.
- When submitting a PR which fixes an issue, please add the following line at the end of the PR's description: `Fixes: #123`, where `#123` is the ID of the issue. By doing so, a link to the PR will appear on the issue's page, and the issue will be automatically closed when the PR is merged (and the merge commit will also contain the issue number, which might be useful when doing some git archaeology).

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_get_next_generation() {
+        let fetcher = setup().await.unwrap();
+
+        let gen = fetcher.fetch_all_generations().await.unwrap();
+
+        let gen1 = fetcher.fetch_next_generation(&gen[0]).await.unwrap();
+        assert!(gen1.is_none());
```
piodul, line 382: It looks like the assertions are swapped in this test - the first invocation of `fetch_next_generation` should return the second generation, while second invocation of `fetch_next_generation` should return `None` because there is no more generations.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+mod tests {
+    use super::*;
+    use scylla::statement::Consistency;
+    use scylla::SessionBuilder;
+
+    const TEST_STREAM_TABLE: &str = "Test.cdc_streams_descriptions_v2";
+    const TEST_GENERATION_TABLE: &str = "Test.cdc_generation_timestamps";
+    const TEST_KEYSPACE: &str = "Test";
+    const GENERATION_1_MILLISECONDS: i64 = 1635882326384;
+    const GENERATION_2_MILLISECONDS: i64 = 1635882224341;
```
piodul, line 182: The first generation's timestamp is larger than the second one. It would make more sense to be the other way around.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+    const TEST_STREAM_1: &str = "0x7fb9f781956cea08c651295720000001";
+    const TEST_STREAM_2: &str = "0x7fc0000000000000c298b9f168000001";
+    const TEST_UUID: &str = "d1dcf782-3d33-11ec-1b45-178f583c09c7";
```
piodul, line 185: Honestly, it's hard to tell by looking at the raw UUID to which generation it is supposed to belong. It would be best to construct a timeuuid in such a way that it is clear by looking at the code to which generation it belongs - for example take a timestamp in the middle between `GENERATION_1_MILLISECONDS` and `GENERATION_2_MILLISECONDS` and construct a timeuuid from it, then check if it belongs to the first generation in a test.

Some documentation links which can be useful:
- Construct a uuid from timestamp: https://docs.rs/uuid/0.8.2/uuid/struct.Uuid.html#method.new_v1
- Construct a timestamp from seconds, nanoseconds and a ClockSequence: https://docs.rs/uuid/0.8.2/uuid/v1/struct.Timestamp.html#method.from_unix
- Construct a Context, which implements ClockSequence: https://docs.rs/uuid/0.8.2/uuid/v1/struct.Context.html#method.new

piodul (changes requested):
There is a problem with the consistency level chosen for the queries. We would like the `GenerationFetcher` to use as high consistency level as possible, however the `system_distributed` keyspace has a replication factor of 3, but the cluster might have less than 3 nodes. Because of that it would be nice if, before each query, the `GenerationFetcher` checked the cluster size and then used appropriate consistency level:

- For 1 node use `Consistency::One`,
- For 2 nodes use `Consistency::Quorum`,
- For 3 nodes use `Consistency::All`.

To get the number of nodes you can send the following query: `SELECT COUNT(*) FROM system.peers`, then add 1 to the returned number. You can use the default consistency level for that query - the `system` keyspace is special and the coordinator which handles the query will only read local data and will not contact other nodes, so the query will work with any cluster size larger than 0.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    // Internal function configurable for testing purposes.
+    pub async fn fetch_generation_by_timeuuid(
+        &self,
+        time: &Uuid,
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
```
Ponewor, line 80: Does this really cover the requirement described as 

> For a given timestamp, which generation contains it

? I'm not sure if a timestamp is the same thing as an Uuid. Also, it kinda makes my work in #4 unfeasible.
@piodul what do you think?

piodul: A timestamp is not the same as a timeuuid, but a timeuuid contains a timestamp + some other parts. This function does not use "the other parts" (because `dateOf` gets rid of them), and it's easier to convert a timeuuid to a timestamp, so I agree that this function should receive a timestamp, not a timeuuid.

kbr-:
> For 3 nodes use Consistency::All.

We should be able to tolerate one failure, therefore we should use Quorum. It gives us the same consistency guarantees as All but has better liveness properties.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+    SELECT streams
+    FROM {}
+    WHERE time = ?;
+    "#,
+            self.streams_table_name
+        )
+    }
+
+    // Internal function configurable for testing purposes.
+    pub async fn fetch_stream_ids(
```
Ponewor, line 132: This should work as well xD

```
pub async fn fetch_stream_ids(
        &self,
        timestamp: &Generation,
    ) -> Result<Vec<StreamID>, Box<dyn Error>> {
        Ok(self.session.query_iter(
            self.get_stream_ids_by_time_query(),
            (Timestamp(timestamp.0), ),
        ).await?.map(|x| StreamID(x.unwrap().into_typed::<(Vec<u8>, )>().unwrap().0)).collect().await)
    }

Comment (macher259) [reply scylla-cdc/src/stream_generations.rs:132]: I don't think so since each row contains `set<blob>` and cql set's equivalent is `Vec<T>` and cql blob's equivalent is `Vec<u8>`, so it ought to be `[...].into_typed::<(Vec<Vec<u8>>, )>()[...]`. I had some problem when trying to implement it, but it could be because of my lack of understanding of Rust.

Comment (Ponewor) [reply scylla-cdc/src/stream_generations.rs:132]: > each row contains `set<blob>`
Would you be so kind to point me to the specific part of documentation supporting this claim?

Comment (macher259) [reply scylla-cdc/src/stream_generations.rs:132]: The table we are querying is `system_distributed.cdc_streams_descriptions_v2`. Using cqlsh we can check its schema using `DESCRIBE`. It gives following output:
```
cqlsh> describe system_distributed.cdc_streams_descriptions_v2;

CREATE TABLE system_distributed.cdc_streams_descriptions_v2 (
    time timestamp,
    range_end bigint,
    streams frozen<set<blob>>,
    PRIMARY KEY (time, range_end)
) WITH CLUSTERING ORDER BY (range_end ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys': 'ALL', 'rows_per_partition': 'ALL'}
    AND comment = ''
    AND compaction = {'class': 'SizeTieredCompactionStrategy'}
    AND compression = {'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1.0
    AND dclocal_read_repair_chance = 0.0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0.0
    AND speculative_retry = '99.0PERCENTILE';
```

Comment (Ponewor) [reply scylla-cdc/src/stream_generations.rs:132]: Yeah, you're right, thank you.

Just for fun: this one works but it's admittedly messier ;)
```rust
pub async fn fetch_stream_ids(
        &self,
        timestamp: &Generation,
    ) -> Result<Vec<StreamID>, Box<dyn Error>> {
        Ok(self.session.query_iter(
            self.get_stream_ids_by_time_query(),
            (Timestamp(timestamp.0), ),
        ).await?.into_typed::<(Vec<Vec<u8>>, )>().flat_map(
            |x| futures::stream::iter(x.unwrap().0.iter().map(|y| StreamID(y.clone())).collect::<Vec<_>>())
        ).collect().await)
    }
```

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,407 @@
...
+                Some(row) => {
+                    let row = row?.0;
+                    Ok(Some(Generation(row)))
+                }
+            };
+        }
+
+        Ok(None)
+    }
+}
```
Ponewor, line 170: Variable shadowing is cool but isn't using one name for four different types a bit too much? At least it's a bit confusing to me, especially because initially what you're refferring to as a singular `row` is actually a `Vec<Row>` (and yeah, I know its length is always 1 in your use cases but it still gives the opposite impression). What do you think @macher259 ? Maybe I am wrong and it's exactly what you're supposed to do in Rust? You probably know better.

I like it this way:
```rust
pub fn return_single_row(rows: Option<Vec<Row>>) -> Result<Option<Generation>, Box<dyn Error>> {
        if let Some(rows) = rows {
            return match rows.into_typed::<(chrono::Duration, )>().next() {
                None => Ok(None),
                Some(row) => {
                    let time = row?.0;
                    Ok(Some(Generation(time)))
                }
            };
        }

        Ok(None)
    }
```

Ponewor: As a side note: you can shorten this to
```rust
pub fn return_single_row(row: Option<Vec<Row>>) -> Result<Option<Generation>, Box<dyn Error>> {
    if let Some(row) = row {
        if let Some(row) = row.into_typed::<(chrono::Duration, )>().next() {
            let row = row?.0;
            return Ok(Some(Generation(row)))
        }
    }

    Ok(None)
}
```

Issue #1, opened 2021-10-29
Set up basic CI


piodul:
We want to have a CI set up which will run for each pull request.

As a start, it would be good to have the following checks:

- `cargo build` - should check that the project builds without errors __and without warnings__,
- `cargo fmt` - so that we enforce a unified formatting style,
- `cargo clippy` - so that the code is more idiomatic and common mistakes are caught automatically,
- `cargo test` - so that we are sure that the tests are passing. Some tests will probably need a running Scylla instance, so the CI should spin up a cluster with two nodes.

It can be based it on scylla-rust-driver's CI: https://github.com/scylladb/scylla-rust-driver/blob/main/.github/workflows/rust.yml

PR #7, opened 2021-11-04
Set up basic workflow

procrastinationfighter:
Fixes: #1 
There are not many changes in comparison to the workflow in the driver's repository, however I tried to do some changes and they were either not necessary or not possible:

- If we want to `build` without warnings, it's better to add a flag (`-D warnings`) to clippy than trying to make `cargo build` fail in case of warnings.
- Creating a Scylla cluster using Github workflows seems really difficult if not impossible. The Github workflows uses Docker and we can't pass the `seeds` argument when creating a container. I described this issue in more detail on Slack. 

I did not use the `cargo check` step, because it was not mentioned in the issue and as far as I know, it only checks for errors that `cargo build` should fail on anyway.

I also removed flags connected to the `examples` directory, since we don't have it yet.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,33 @@
+name: Rust
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    services:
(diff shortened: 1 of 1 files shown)


piodul (approved):
Looks good. Before merging, please do the following:

- Squash the commits into one and force push,
- Add a `Fixes: #1` line to the PR description so that the corresponding issue is closed when this PR is merged.

PR #8, opened 2021-11-07
Generation Fetcher implementation.

macher259:
Fixes: #2 

I have deleted init commit by accident while squashing my commits. This resulted in the closing of the previous PR. I couldn't reopen it, so I am opening another one. >_<

Applied changes suggested by #6

### DIFF
--- a/scylla-cdc/src/stream_generations.rs
+++ b/scylla-cdc/src/stream_generations.rs
@@ -1,6 +1,6 @@
 use futures::stream::StreamExt;
 use scylla::batch::Consistency;
-use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromCqlValError::BadCqlType, FromRow};
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromCqlValError::BadCqlType};
 use scylla::frame::response::result::{CqlValue, Row};
 use scylla::frame::value::{Timestamp, Value, ValueTooBig};
 use scylla::query::Query;
--- a/scylla-cdc/src/stream_generations.rs
+++ b/scylla-cdc/src/stream_generations.rs
@@ -1,6 +1,6 @@
 use futures::stream::StreamExt;
 use scylla::batch::Consistency;
-use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromCqlValError::BadCqlType};
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError};
 use scylla::frame::response::result::{CqlValue, Row};
 use scylla::frame::value::{Timestamp, Value, ValueTooBig};
 use scylla::query::Query;
@@ -32,7 +32,10 @@ impl Value for StreamID {
 
 impl FromCqlVal<CqlValue> for StreamID {
     fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
-        let id = cql_val.as_blob().ok_or(BadCqlType)?.to_owned();
+        let id = cql_val
+            .as_blob()
+            .ok_or(FromCqlValError::BadCqlType)?
+            .to_owned();
         Ok(StreamID { id })
     }
 }
@@ -165,10 +168,11 @@ impl GenerationFetcher {
     }
 
     /// Given a generation return identifiers of all streams of this generation.
+    /// Streams are grouped by vnodes.
     pub async fn fetch_stream_ids(
         &self,
         generation: &GenerationTimestamp,
-    ) -> Result<Vec<StreamID>, Box<dyn Error>> {
+    ) -> Result<Vec<Vec<StreamID>>, Box<dyn Error>> {
         let mut result_vec = Vec::new();
 
         let mut query =
@@ -184,10 +188,9 @@ impl GenerationFetcher {
 
         while let Some(next_row) = rows.next().await {
             let (ids,) = next_row?;
-            for id in ids {
-                result_vec.push(id);
-            }
+            result_vec.push(ids);
         }
+
         Ok(result_vec)
     }
 
@@ -486,10 +489,15 @@ mod tests {
 
         let stream_ids = fetcher.fetch_stream_ids(&gen).await.unwrap();
 
-        let correct_stream_ids: Vec<StreamID> = vec![TEST_STREAM_1, TEST_STREAM_2]
+        let correct_stream_ids: Vec<Vec<StreamID>> = vec![[TEST_STREAM_1, TEST_STREAM_2]]
             .iter()
-            .map(|stream| StreamID {
-                id: hex::decode(stream.strip_prefix("0x").unwrap()).unwrap(),
+            .map(|stream_vec| {
+                stream_vec
+                    .iter()
+                    .map(|stream| StreamID {
+                        id: hex::decode(stream.strip_prefix("0x").unwrap()).unwrap(),
+                    })
+                    .collect()
             })
             .collect();
 
--- a/scylla-cdc/src/stream_generations.rs
+++ b/scylla-cdc/src/stream_generations.rs
@@ -1,18 +1,41 @@
 use futures::stream::StreamExt;
 use scylla::batch::Consistency;
-use scylla::frame::response::result::Row;
-use scylla::frame::value::Timestamp;
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromCqlValError::BadCqlType, FromRow};
+use scylla::frame::response::result::{CqlValue, Row};
+use scylla::frame::value::{Timestamp, Value, ValueTooBig};
 use scylla::query::Query;
-use scylla::{IntoTypedRows, Session};
+use scylla::{FromRow, IntoTypedRows, Session};
 use std::error::Error;
 use std::sync::Arc;
-use uuid::Uuid;
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub struct Generation(chrono::Duration);
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
+pub struct GenerationTimestamp {
+    timestamp: chrono::Duration,
+}
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub struct StreamID(Vec<u8>);
+impl Value for GenerationTimestamp {
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
+        Timestamp(self.timestamp).serialize(buf)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
+pub struct StreamID {
+    id: Vec<u8>,
+}
+
+impl Value for StreamID {
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
+        self.id.serialize(buf)
+    }
+}
+
+impl FromCqlVal<CqlValue> for StreamID {
+    fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
+        let id = cql_val.as_blob().ok_or(BadCqlType)?.to_owned();
+        Ok(StreamID { id })
+    }
+}
 
 /// Component responsible for managing stream generations.
 pub struct GenerationFetcher {
@@ -25,7 +48,6 @@ pub struct GenerationFetcher {
 const DEFAULT_PAGE_SIZE: i32 = 5000;
 
 impl GenerationFetcher {
-    /// Creates new instance of GenerationFetcher.
     pub fn new(session: &Arc<Session>) -> GenerationFetcher {
         GenerationFetcher {
             generations_table_name: "system_distributed.cdc_generation_timestamps".to_string(),
@@ -37,33 +59,33 @@ impl GenerationFetcher {
     // Function instead of constant for testing purposes.
     fn get_all_stream_generations_query(&self) -> String {
         format!(
-            r#"
+            "
     SELECT time
     FROM {}
     WHERE key = 'timestamps';
-    "#,
+    ",
             self.generations_table_name
         )
     }
 
     /// In case of a success returns a vector containing all the generations in the database.
     /// Propagates all errors.
-    pub async fn fetch_all_generations(&self) -> Result<Vec<Generation>, Box<dyn Error>> {
-        let mut query = Query::new(self.get_all_stream_generations_query());
-        select_consistency(&self.session, &mut query).await?;
+    pub async fn fetch_all_generations(&self) -> Result<Vec<GenerationTimestamp>, Box<dyn Error>> {
+        let mut generations = Vec::new();
+
+        let mut query =
+            new_distributed_system_query(self.get_all_stream_generations_query(), &self.session)
+                .await?;
         query.set_page_size(DEFAULT_PAGE_SIZE);
 
         let mut rows = self
             .session
             .query_iter(query, &[])
             .await?
-            .into_typed::<(chrono::Duration,)>();
+            .into_typed::<GenerationTimestamp>();
 
-        let mut generations = vec![];
-
-        while let Some(next_row) = rows.next().await {
-            let (generation,) = next_row?;
-            generations.push(Generation(generation));
+        while let Some(generation) = rows.next().await {
+            generations.push(generation?)
         }
 
         Ok(generations)
@@ -72,14 +94,14 @@ impl GenerationFetcher {
     // Function instead of constant for testing purposes.
     fn get_generation_by_timestamp_query(&self) -> String {
         format!(
-            r#"
+            "
     SELECT time
     FROM {}
     WHERE key = 'timestamps'
     AND time <= ?
     ORDER BY time DESC
     LIMIT 1;
-    "#,
+    ",
             self.generations_table_name
         )
     }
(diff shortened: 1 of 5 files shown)


scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,531 @@
...
+        ] {
+            session
+                .query(format!("TRUNCATE {};", table), &[])
+                .await
+                .unwrap();
+        }
+    }
+
+    // Populate test tables with given data.
+    async fn populate_test_db(session: &Session) {
```
piodul, line 345: This function no longer works with a one node cluster because `.query` uses CL=QUORUM by default and the test keyspace now uses RF=3, so at least two nodes are needed for such query to work. I suggest extracting the `get_cluster_size` to a standalone function and use it both in `GenerationFetcher` and the tests.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    // Function instead of constant for testing purposes.
+    fn get_generation_by_timestamp_query(&self) -> String {
+        format!(
+            r#"
+    SELECT time
+    FROM {}
+    WHERE key = 'timestamps'
+    AND time <= ?
```
piodul, line 77: Please add an explicit `ORDER BY time DESC` here. When creating a new table without specifying the ordering of clustering key, the clustering key is ordered ASC by default - the generations table is ordered DESC, so an explicit mention of ordering should reduce confusion.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub async fn fetch_generation_by_timeuuid(
+        &self,
+        time: &Uuid,
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_generation_by_timeuuid_query());
+        self.select_consistency(&mut query).await?;
+
+        let result = self.session.query(query, (time,)).await?.rows;
+
+        GenerationFetcher::return_single_row(result)
```
piodul, line 121: Let's not duplicate the logic of `fetch_generation_by_timestamp`. You can extract a timestamp from the timeuuid and pass it to the `fetch_generation_by_timestamp` function.

I think the extraction can be done in the following way:

```rust
let (secs, nanos) = time
    .to_timestamp()
    .ok_or("Uuid is not a timeuuid")?
    .to_unix();
let timestamp =
    chrono::Duration::seconds(secs as i64) + chrono::Duration::nanoseconds(nanos as i64);
```

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,531 @@
...
+        let fetcher = setup().await.unwrap();
+
+        let correct_gen = vec![
+            Generation(chrono::Duration::milliseconds(GENERATION_NEW_MILLISECONDS)),
+            Generation(chrono::Duration::milliseconds(GENERATION_OLD_MILLISECONDS)),
+        ];
+
+        let gen = fetcher.fetch_all_generations().await.unwrap();
+
+        assert_eq!(gen.len(), correct_gen.len());
```
piodul, line 398: Nit: no need to compare lengths of `gen` and `correct_gen` if you are comparing the whole values later (`assert_eq!(gen, correct_gen)`).

piodul (changes requested):

piodul (approved):
LGTM

@Ponewor your task depends on this functionality, could you review this PR, too?

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+    generations_table_name: String,
+    streams_table_name: String,
+    session: Arc<Session>,
+}
+
+// Number taken from: https://www.scylladb.com/2017/11/17/7-rules-planning-queries-maximum-performance/.
+const DEFAULT_PAGE_SIZE: i32 = 5000;
+
+impl GenerationFetcher {
+    /// Creates new instance of GenerationFetcher.
```
kbr-, line 28: Redundant comment

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+use scylla::frame::response::result::Row;
+use scylla::frame::value::Timestamp;
+use scylla::query::Query;
+use scylla::{IntoTypedRows, Session};
+use std::error::Error;
+use std::sync::Arc;
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Generation(chrono::Duration);
```
kbr-, line 12: A generation consists of a "generation timestamp" - which is a point in time when the generation starts operating - and a set of streams (stream IDs). The streams are part of the generation. This is just the timestamp, so perhaps `GenerationTimestamp` would be a better name?

Also, since the generation timestamp *is* a point in time, representing it using a `Duration` is probably not a good idea. If you want to represent a point in time using a `Duration` you also need to specify the epoch (the time point is then denoted by the epoch + the duration). A better type would probably be `Instant`.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+                return Ok(Some(Generation(time)));
+            }
+        }
+
+        Ok(None)
+    }
+}
+
+// Returns current cluster size in case of a success.
+async fn get_cluster_size(session: &Session) -> Result<usize, Box<dyn Error>> {
```
kbr-, line 205: Doesn't the driver store the topology knowledge? Can't the cluster size be obtained directly from `session`?
(it's still good to ensure that we have sufficiently fresh topology information, but it should be also possible from `session` to explicitly refresh it)

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+    "#,
+            self.generations_table_name
+        )
+    }
+
+    /// In case of a success returns a vector containing all the generations in the database.
+    /// Propagates all errors.
+    pub async fn fetch_all_generations(&self) -> Result<Vec<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_all_stream_generations_query());
+        select_consistency(&self.session, &mut query).await?;
```
kbr-, line 53: How about a function

```
async fn new_distributed_system_query(stmt: String, session: &Session) -> Result<Query, Box<dyn Error>>
```

which implements this pattern (`Query::new` + `select_consistency`)? This `select_consistency` function is only used in such scenarios.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+    ) -> Result<Option<Generation>, Box<dyn Error>> {
+        let mut query = Query::new(self.get_generation_by_timestamp_query());
+        select_consistency(&self.session, &mut query).await?;
+
+        let result = self.session.query(query, (Timestamp(*time),)).await?.rows;
+
+        GenerationFetcher::return_single_row(result)
+    }
+
+    /// Given a timeuuid of a ti of an operation fetch generation that was operating when this operation was performed.
```
kbr-, line 102: "timeuuid of a ti"?

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// If no such generation exists, returns None.
+    /// Propagates errors.
+    pub async fn fetch_generation_by_timeuuid(
```
kbr-, line 105: Do we want to have this function in `GenerationFetcher`'s interface? It doesn't really add any power to this class, it doesn't need any access to `GenerationFetcher`'s internals. Why not leave it to the user of `GenerationFetcher` to extract the timestamps from their timeuuids on their own and just use `fetch_generation_by_timestamp`?

(and if they want to do this often they can easily implement `async fn fetch_generation_by_timeuuid(&GenerationFetcher, &Uuid)`)

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+
+    // Constructs mock table with the same schema as the original one's.
+    fn construct_generation_table_query() -> String {
+        format!(
+            "CREATE TABLE IF NOT EXISTS {}(
+    key text,
+    time timestamp,
+    expired timestamp,
+    PRIMARY KEY (key, time)
+) WITH CLUSTERING ORDER BY (time DESC)
```
kbr-, line 264: every setting below this line is irrelevant, the test would be nicer if the schema definition was as short as possible

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+
+    // Constructs mock table with the same schema as the original one's.
+    fn construct_stream_table_query() -> String {
+        format!(
+            "CREATE TABLE IF NOT EXISTS {} (
+    time timestamp,
+    range_end bigint,
+    streams frozen<set<blob>>,
+    PRIMARY KEY (time, range_end)
+) WITH CLUSTERING ORDER BY (range_end ASC)
```
kbr-, line 291: every setting below this line is irrelevant, the test would be nicer if the schema definition was as short as possible

kbr- (commented):
The commit message should describe what the commit does.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+    SELECT streams
+    FROM {}
+    WHERE time = ?;
+    "#,
+            self.streams_table_name
+        )
+    }
+
+    /// Given a generation return identifiers of all streams of this generation.
+    pub async fn fetch_stream_ids(
```
Ponewor, line 166: This function and `fetch_all_generations` are almost identical - they do the same thing: prepare a query, execute it, iterate over selected rows and save some value to the vector, return the vector. One could try to abstract this behaviour away but it's probably impossible since required `FromRow` trait is private (not to mention the fact that this may be an overkill for just the two calls). For aesthetical reasons you could at least try to keep the implementations as similar as possible. It's mostly about being consistent in where you put the empty vector initialization - before all the query business or after it.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+        const CONTEXT_COUNT: u16 = 2137; // Just a random number.
+        let context = Context::new(CONTEXT_COUNT);
+
+        let time = std::time::Duration::from_millis(milliseconds as u64);
+        let timestamp = Timestamp::from_unix(context, time.as_secs(), time.subsec_nanos());
+
+        let u = Uuid::new_v1(timestamp, &[1, 2, 3, 4, 5, 6])
+            .expect("Couldn't create a valid uuid value.");
+        println!("{}", u);
+        u
```
Ponewor, line 409: What is it?

piodul: The last expression in a function serves as its return value - you do not need to write `return u;`. Here `u` is assigned to a variable in order to be printed out, then it is returned.

scylla-cdc/src/stream_generations.rs
piodul: > Doesn't the driver store the topology knowledge? Can't the cluster size be obtained directly from `session`?

The driver internally keeps information about the topology, but does not expose it to the users. I agree it would be good if it was available to users.

> (it's still good to ensure that we have sufficiently fresh topology information, but it should be also possible from `session` to explicitly refresh it)

Topology refresh is quite an expensive operation because it reads various details about all the nodes in the cluster (e.g. their tokens, dcs) and creates a relatively heavyweight object to keep that information. Here, we are interested in the cluster size only, so I think it's OK if this function stays in its current form.

Ponewor:
The other day I was trying to run the tests and one of them unexpectedly failed. This is the message that I got:
```
called `Result::unwrap()` on an `Err` value: DbError(Invalid, "unconfigured table cdc_streams_descriptions_v2")
thread 'stream_generation::tests::test_fetch_all_generations' panicked at 'called `Result::unwrap()` on an `Err` value: DbError(Invalid, "unconfigured table cdc_streams_descriptions_v2")', scylla-cdc/src/stream_generation.rs:412:59
stack backtrace:
```
After reruninng the tests everything worked fine as it did during all the earlier runs so it seems like a completely random incident. Does anyone know what's happening here?

piodul:
> The other day I was trying to run the tests and one of them unexpectedly failed. This is the message that I got:
> 
> ```
> called `Result::unwrap()` on an `Err` value: DbError(Invalid, "unconfigured table cdc_streams_descriptions_v2")
> thread 'stream_generation::tests::test_fetch_all_generations' panicked at 'called `Result::unwrap()` on an `Err` value: DbError(Invalid, "unconfigured table cdc_streams_descriptions_v2")', scylla-cdc/src/stream_generation.rs:412:59
> stack backtrace:
> ```
> 
> After reruninng the tests everything worked fine as it did during all preceding runs so it seems like a completely random incident. Does anyone know what's happening here?

@Ponewor I think this is because the test doesn't wait for schema agreement after a table is created. Schema changes (i.e. creating/deleting tables and keyspaces) are asynchronous - it is possible to create a table on one node, then try to insert to that table on another node and fail because the other node doesn't know about the table yet.

@macher259 After creating the keyspace and after creating the tables (so in two situations) you should wait for schema agreement before moving to the actual test. There is a method in the Session made specifically for that - [`Session::await_schema_agreement`](https://docs.rs/scylla/0.3.0/scylla/transport/session/struct.Session.html#method.await_schema_agreement).

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,540 @@
...
+use scylla::{IntoTypedRows, Session};
+use std::error::Error;
+use std::sync::Arc;
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Generation(chrono::Duration);
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct StreamID(Vec<u8>);
```
Ponewor, line 15: Can we turn this into
```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
pub struct StreamID{id: Vec<u8>}

impl Value for StreamID {
    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
        self.id.serialize(buf)
    }
}
```
this? It would make using this type in other modules easier.

scylla-cdc/src/stream_generations.rs
macher259: Ad 2:
I am not sure about that. In `std::time::Instant`'s docs it is said that:
`Instants are opaque types that can only be compared to one another. There is no method to get “the number of seconds” from an instant. Instead, it only allows measuring the duration between two instants (or comparing two instants).`

I may be mistaken, but I think using `Duration` is better since Scylla Rust Driver uses it internally for representing `Time` and `Timestamp`. 
And... I am not sure how to implement this using `Instant`.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,508 @@
+use futures::stream::StreamExt;
+use scylla::batch::Consistency;
+use scylla::cql_to_rust::FromCqlValError::BadCqlType;
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromRow};
```
Ponewor, line 4: Please merge these imports.

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,507 @@
...
+        let mut rows = self
+            .session
+            .query_iter(query, &[])
+            .await?
+            .into_typed::<(GenerationTimestamp,)>();
+
+        while let Some(next_row) = rows.next().await {
+            let (generation,) = next_row?;
+            generations.push(generation);
+        }
```
Ponewor, line 97:
```rust
    .into_typed::<GenerationTimestamp>();

while let Some(generation) = rows.next().await {
    generations.push(generation?)
}
```

Ponewor (changes requested):

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    // Return single row containing generation.
+    fn return_single_row(
+        row: Option<Vec<Row>>,
+    ) -> Result<Option<GenerationTimestamp>, Box<dyn Error>> {
+        if let Some(row) = row {
+            if let Some(row) = row.into_typed::<(GenerationTimestamp,)>().next() {
+                let generation = row?.0;
+                return Ok(Some(generation));
```
Ponewor, line 209:
```rust
if let Some(generation) = row.into_typed::<GenerationTimestamp>().next() {
    return Ok(Some(generation?));
```

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,507 @@
...
+        Timestamp(self.timestamp).serialize(buf)
+    }
+}
+
+impl FromCqlVal<CqlValue> for GenerationTimestamp {
+    fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
+        let timestamp = cql_val.as_duration().ok_or(BadCqlType)?;
+        Ok(GenerationTimestamp { timestamp })
+    }
+}
```
Ponewor, line 27: When you'll be done with the last two of my nitpicks this code will be no longer needed.

macher259:
Now fixes: #11

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,498 @@
...
+            if let Some(generation) = row.into_typed::<GenerationTimestamp>().next() {
+                return Ok(Some(generation?));
+            }
+        }
+
+        Ok(None)
+    }
+}
+
+// Returns current cluster size in case of a success.
```
Ponewor, line 208: Some of your functions are documented with `//` and some with `///`. Can we use `///` everywhere?

scylla-cdc/src/stream_generations.rs
```diff
@@ -0,0 +1,498 @@
...
+use scylla::batch::Consistency;
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError, FromCqlValError::BadCqlType};
+use scylla::frame::response::result::{CqlValue, Row};
+use scylla::frame::value::{Timestamp, Value, ValueTooBig};
+use scylla::query::Query;
+use scylla::{FromRow, IntoTypedRows, Session};
+use std::error::Error;
+use std::sync::Arc;
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
```
Ponewor, line 11: Is this working? Aren't you missing `use scylla::cql_to_rust::FromRow`?

scylla-cdc/src/stream_generations.rs
macher259: I don't think thats a good idea since /// comments  are the ones that are parsed when creating docs.

scylla-cdc/src/stream_generations.rs
macher259: I have `use scylla::FromRow` below.
It certainly works on a new version of the driver for me. 
Maybe *they* re-exported this struct in the new version.

Ponewor: To make it work I have to import both of them (and they do it the same way in the docs iirc).

macher259: Are you sure that you are using correct scylla Version?
In any case it passes worlflow check from github actions on my fork.

Ponewor: Right, it was the version, thanks.

scylla-cdc/src/stream_generations.rs
Ponewor: So your intention is to create doc comments only for public functions?

macher259: Yes, I think it is a good idea to differentiate between comments intended to be read by a dev working on this repo or one that wants to know internal details by reading source code and ones intended for devs that want only to use exposed api.

Ponewor: As you wish :)

Ponewor:
LGTM

kbr- (approved):
LGTM.

The commits "Removed unneeded method fetch_generation_by_timeuuid..." and "Changed api to returning stream ids grouped by vnodes. " are really amends to the "Generation Fetcher implementation." commit, so they could be squashed to that commit (in general: changes made to a certain commit during review can be squashed into that commit). But I guess they don't hurt, they have nice descriptions so we can leave them.

piodul (approved):

Issue #3, opened 2021-10-29
Basic stream reader component


piodul:
We should have a component which, given a set of streams and a CDC log table name, periodically fetches changes (data from CDC log) from those streams.

Requirements:
- When constructed, it should take the following configuration:
    - A _set_ of stream IDs to track changes from,
    - A timestamp from which the reader should start reading the streams,
    - Time interval parameters such as: query time window size, safety interval, interval to sleep after reading a time window.
- The reader should run an asynchronous process which follows an algorithm similar to this:
    - Keep track of timestamp `T`,
    - Choose the time window end `U = max(T, min(T + window_size, now - safety_interval))`,
    - Query rows from the time interval `[T, U)`
        - You should read rows from all streams at once, in a single query - you can use the `WHERE "cdc$stream_id" IN (?, ?, ...)` notation to fetch rows from multiple streams at once.
    - Process rows returned from the query,
    - Sleep for the configured amount of time,
    - Repeat.
- For now, the reader should only print received changes (in the future, the data will be passed to something like a user-defined callback).
- It should be possible to set a timestamp up to which the reader is supposed to read data. It should be possible to do it while the fetcher is running, or before it starts running.
- After reading data from the CDC log up to given timestamp, the asynchronous process should stop.

PR #9, opened 2021-11-09
Added Stream Reader Component

Gor027:
# Fixes: #3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /target
 Cargo.lock
 .idea/
+
--- a/scylla-cdc/src/cdc_types.rs
+++ b/scylla-cdc/src/cdc_types.rs
@@ -1,0 +1,42 @@
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError};
+use scylla::frame::response::result::CqlValue;
+use scylla::frame::value::{Timestamp, Value, ValueTooBig};
+use scylla::FromRow;
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
+pub struct GenerationTimestamp {
+    pub(crate) timestamp: chrono::Duration,
+}
+
+impl Value for GenerationTimestamp {
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
+        Timestamp(self.timestamp).serialize(buf)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, FromRow)]
+pub struct StreamID {
+    pub(crate) id: Vec<u8>,
+}
+
+impl Value for StreamID {
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
+        self.id.serialize(buf)
+    }
+}
+
+impl FromCqlVal<CqlValue> for StreamID {
+    fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
+        let id = cql_val
+            .as_blob()
+            .ok_or(FromCqlValError::BadCqlType)?
+            .to_owned();
+        Ok(StreamID { id })
+    }
+}
+
+impl StreamID {
+    pub fn new(stream_id: Vec<u8>) -> StreamID {
+        StreamID { id: stream_id }
+    }
+}
--- a/scylla-cdc/src/lib.rs
+++ b/scylla-cdc/src/lib.rs
@@ -1,4 +1,6 @@
+mod cdc_types;
 pub mod consumer;
+pub mod reader;
 pub mod stream_generations;
 
 #[cfg(test)]
--- a/scylla-cdc/src/reader.rs
+++ b/scylla-cdc/src/reader.rs
@@ -1,0 +1,417 @@
+use std::cmp::{max, min};
+use std::sync::Arc;
+use std::time;
+
+use chrono;
+use futures::StreamExt;
+use scylla::frame::response::result::Row;
+use scylla::frame::value::Timestamp;
+use scylla::Session;
+use tokio::time::sleep;
+
+use crate::cdc_types::StreamID;
+
+pub struct StreamReader {
+    session: Arc<Session>,
+    stream_id_vec: Vec<StreamID>,
+    lower_timestamp: chrono::Duration,
+    window_size: chrono::Duration,
+    safety_interval: chrono::Duration,
+    upper_timestamp: tokio::sync::Mutex<Option<chrono::Duration>>,
+    sleep_interval: time::Duration,
+}
+
+impl StreamReader {
+    pub fn new(
+        session: &Arc<Session>,
+        stream_ids: Vec<StreamID>,
+        start_timestamp: chrono::Duration,
+        window_size: chrono::Duration,
+        safety_interval: chrono::Duration,
+        sleep_interval: time::Duration,
+    ) -> StreamReader {
+        StreamReader {
+            session: session.clone(),
+            stream_id_vec: stream_ids,
+            lower_timestamp: start_timestamp,
+            window_size,
+            safety_interval,
+            upper_timestamp: Default::default(),
+            sleep_interval,
+        }
+    }
+
+    pub async fn set_upper_timestamp(&self, new_upper_timestamp: chrono::Duration) {
+        let mut guard = self.upper_timestamp.lock().await;
+        *guard = Some(new_upper_timestamp);
+    }
+
+    pub async fn fetch_cdc(
+        &self,
+        keyspace: String,
+        table_name: String,
+        mut after_fetch_callback: impl FnMut(Row),
+    ) -> anyhow::Result<()> {
+        let query = format!(
+            "SELECT * FROM {}.{}_scylla_cdc_log \
+            WHERE \"cdc$stream_id\" in ? \
+            AND \"cdc$time\" >= minTimeuuid(?) \
+            AND \"cdc$time\" < maxTimeuuid(?)",
+            keyspace, table_name
+        );
+        let query_base = self.session.prepare(query).await?;
+        let mut window_begin = self.lower_timestamp;
+
+        loop {
+            let now = chrono::Local::now().timestamp_millis();
+
+            let window_end = max(
+                window_begin,
+                min(
+                    window_begin + self.window_size,
+                    chrono::Duration::milliseconds(now - self.safety_interval.num_milliseconds()),
+                ),
+            );
+
+            let mut rows_stream = self
+                .session
+                .execute_iter(
+                    query_base.clone(),
+                    (
+                        &self.stream_id_vec,
+                        Timestamp(window_begin),
+                        Timestamp(window_end),
+                    ),
+                )
+                .await?;
+
+            while let Some(row) = rows_stream.next().await {
+                after_fetch_callback(row?);
+            }
+
+            if let Some(timestamp_to_stop) = self.upper_timestamp.lock().await.as_ref() {
+                if window_end >= *timestamp_to_stop {
+                    break;
+                }
+            }
+
+            window_begin = window_end;
+            sleep(self.sleep_interval).await;
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use futures::stream::StreamExt;
+    use scylla::batch::Consistency;
+    use scylla::query::Query;
+    use scylla::SessionBuilder;
+
+    use super::*;
+
+    const SECOND_IN_MICRO: i64 = 1_000_000;
+    const SECOND_IN_MILLIS: i64 = 1_000;
+    const TEST_KEYSPACE: &str = "test";
+    const TEST_TABLE: &str = "t";
+    const SLEEP_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
+    const WINDOW_SIZE: i64 = SECOND_IN_MILLIS / 10 * 3;
+    const SAFETY_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
+    const START_TIME_DELAY_IN_MILLIS: i64 = 2 * SECOND_IN_MILLIS;
+
+    impl StreamReader {
+        fn test_new(
+            session: &Arc<Session>,
+            stream_ids: Vec<StreamID>,
+            start_timestamp: chrono::Duration,
+            window_size: chrono::Duration,
+            safety_interval: chrono::Duration,
+            sleep_interval: time::Duration,
+        ) -> StreamReader {
+            StreamReader {
+                session: session.clone(),
+                stream_id_vec: stream_ids,
+                lower_timestamp: start_timestamp,
+                window_size,
+                safety_interval,
+                upper_timestamp: Default::default(),
+                sleep_interval,
+            }
+        }
+    }
+
+    async fn get_test_stream_reader(session: &Arc<Session>) -> anyhow::Result<StreamReader> {
+        let stream_id_vec = get_cdc_stream_id(session).await?;
+
+        let start_timestamp: chrono::Duration = chrono::Duration::milliseconds(
+            chrono::Local::now().timestamp_millis() - START_TIME_DELAY_IN_MILLIS,
+        );
+        let sleep_interval: time::Duration = time::Duration::from_millis(SLEEP_INTERVAL as u64);
+        let window_size: chrono::Duration = chrono::Duration::milliseconds(WINDOW_SIZE);
+        let safety_interval: chrono::Duration = chrono::Duration::milliseconds(SAFETY_INTERVAL);
+
+        let reader = StreamReader::test_new(
+            session,
+            stream_id_vec,
+            start_timestamp,
+            window_size,
+            safety_interval,
+            sleep_interval,
+        );
+
+        Ok(reader)
+    }
+
+    fn get_create_table_query() -> String {
+        format!("CREATE TABLE IF NOT EXISTS {}.{} (pk int, t int, v text, s text, PRIMARY KEY (pk, t)) WITH cdc = {{'enabled':true}};",
+                TEST_KEYSPACE,
+                TEST_TABLE
+        )
+    }
+
+    async fn create_test_db(session: &Arc<Session>) -> anyhow::Result<()> {
+        let mut create_keyspace_query = Query::new(format!(
+            "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class': 'SimpleStrategy', 'replication_factor': 1}};",
+            TEST_KEYSPACE
+        ));
+        create_keyspace_query.set_consistency(Consistency::All);
+
+        session.query(create_keyspace_query, &[]).await?;
+        session.await_schema_agreement().await?;
+
+        // Create test table
+        let create_table_query = get_create_table_query();
+        session.query(create_table_query, &[]).await?;
+        session.await_schema_agreement().await?;
+
+        let table_name_under_keyspace = format!("{}.{}", TEST_KEYSPACE, TEST_TABLE);
+        session
+            .query(format!("TRUNCATE {};", table_name_under_keyspace), &[])
+            .await?;
+        session
+            .query(
+                format!("TRUNCATE {}_scylla_cdc_log;", table_name_under_keyspace),
+                &[],
+            )
+            .await?;
+        Ok(())
+    }
+
+    async fn populate_db_with_pk(session: &Arc<Session>, pk: u32) -> anyhow::Result<()> {
+        let table_name_under_keyspace = format!("{}.{}", TEST_KEYSPACE, TEST_TABLE);
+        for i in 0..3 {
+            session
+                .query(
+                    format!(
+                        "INSERT INTO {} (pk, t, v, s) VALUES ({}, {}, 'val{}', 'static{}');",
+                        table_name_under_keyspace, pk, i, i, i
+                    ),
+                    &[],
+                )
+                .await?;
+        }
+        session.await_schema_agreement().await?;
+
+        Ok(())
+    }
+
+    async fn get_cdc_stream_id(session: &Arc<Session>) -> anyhow::Result<Vec<StreamID>> {
+        let table_name_under_keyspace = format!("{}.{}", TEST_KEYSPACE, TEST_TABLE);
+        let query_stream_id = format!(
+            "SELECT DISTINCT \"cdc$stream_id\" FROM {}_scylla_cdc_log;",
+            table_name_under_keyspace
+        );
+
+        let mut rows = session
+            .query_iter(query_stream_id, ())
+            .await?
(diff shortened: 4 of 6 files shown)


scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub fn set_lower_timestamp(&mut self, new_lower_timestamp: chrono::Duration) {
+        self.upper_timestamp = Some(new_lower_timestamp);
+    }
+
+    pub async fn fetch_cdc(
+        &self,
+        keyspace: String,
+        table_name: &str,
+    ) -> Result<(), Box<dyn Error>> {
+        const URI: &str = "172.17.0.2:9042";
```
Ponewor, line 55: This must not be fixed like this. Your tests pass just because incidentally your test database's URI matches this one.
This address should be passed to `Reader` through `new` or even better - entire `Session` object should be dependency injected.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    async fn create_test_db(session: &Session) {
+        let mut create_keyspace_query = Query::new(format!(
+            "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class': 'SimpleStrategy', 'replication_factor': 1}};",
+            TEST_KEYSPACE
+        ));
+        create_keyspace_query.set_consistency(Consistency::All);
+        session.query(create_keyspace_query, &[]).await.unwrap();
+
+        // Create test table
+        let create_table_query = get_create_table_query();
```
Ponewor, line 189: This will fail if the table already exists. You should add `IF NOT EXISTS` to your query or make sure you tear everything down after the tests are run.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+                ),
+                (),
+            )
+            .await
+            .unwrap();
+    }
+
+    async fn get_cdc_stream_id(session: &Session) -> Result<StreamID, Box<dyn Error>> {
+        let query_stream_id = format!(
+            "SELECT \"cdc$stream_id\" FROM {}_scylla_cdc_log limit 1;",
```
Ponewor, line 232: Capitalize `limit` keyword for consistency.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+        let mut stream_id_vec: StreamID = vec![];
+        let mut rows = session
+            .query_iter(query_stream_id, &[])
+            .await?
+            .into_typed::<(StreamID,)>();
+
+        while let Some(next_row) = rows.next().await {
+            stream_id_vec = next_row.unwrap().0;
+        }
+        Ok(stream_id_vec)
```
Ponewor, line 245: Are you sure you want to loop here?
`Ok(rows.next().await.unwrap()?.0)` should do the trick as well

Ponewor (changes requested):

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+            WHERE \"cdc$stream_id\" in (?) \
+            AND \"cdc$time\" >= maxTimeuuid(?) \
+            AND \"cdc$time\" < minTimeuuid(?)",
+            keyspace, table_name
+        );
+        let query_base = session.prepare(query).await.unwrap();
+        let mut continue_fetch = true;
+        let mut lower_timestamp = self.lower_timestamp;
+
+        while continue_fetch {
```
Ponewor, line 68: Why not just `loop{}` and `break` when needed?

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    // Mutable access.
+    fn lower_timestamp_mut(&mut self) -> Option<chrono::Duration> {
+        self.upper_timestamp
+    }
+
+    pub fn set_lower_timestamp(&mut self, new_lower_timestamp: chrono::Duration) {
+        self.upper_timestamp = Some(new_lower_timestamp);
+    }
```
Ponewor, line 48: Are those functions properly named? Because they mention lower_timestamp in their names but they deal with upper_timestamp.

Ponewor: Those functions are not used anywhere. If they're part of the required interface could you add some tests for them?

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+                None => {
+                    upper_timestamp = max(
+                        lower_timestamp,
+                        min(
+                            lower_timestamp + self.window_size,
+                            now - self.safety_interval,
+                        ),
+                    )
+                }
+            }
```
Ponewor, line 88: Or maybe
```rust
let upper_timestamp = self.upper_timestamp.unwrap_or(max(
    lower_timestamp,
    min(
        lower_timestamp + self.window_size,
        now - self.safety_interval,
     ),
));
```

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+        let mut continue_fetch = true;
+        let mut lower_timestamp = self.lower_timestamp;
+
+        while continue_fetch {
+            let now = chrono::Duration::from_std(
+                SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap(),
+            )
+            .unwrap();
```
Ponewor, line 74: Is it just me or does this linebreak look weird?

Ponewor (changes requested):

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    async fn delete_test_data_in_db(session: &Session) {
+        session
+            .query(format!("DROP KEYSPACE IF EXISTS {}", TEST_KEYSPACE), ())
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    async fn fetch_cdc_data() -> Result<(), Box<dyn std::error::Error>> {
+        let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "172.17.0.2:9042".to_string());
```
Ponewor, line 257: Can we standardise the default value for this URI or not use any at all? @macher259 uses different one.

scylla-cdc/src/reader.rs
Ponewor: To clarify: `delete_test_data_in_db` is not called when your tests fail (for example when some database or networking error happens - simply provide invalid database URI and you'll experience it) leaving the test database in a messy state.

scylla-cdc/src/reader.rs
piodul, line 48: While writing the issue's description I might have been imprecise in phrasing the point about setting the end timestamp for the reader - I meant that both options should be possible: setting the timestamp before the reader is running AND while it is running. However, the current API does not allow for that because of Rust's ownership rules.

- `Reader::fetch_cdc` which runs the reading process borrows `Reader` by immutable reference (`&self`),
- `Reader::set_lower_timestamp` borrows `Reader` by mutable reference (`&mut self`).

Because Rust does not allow a mutable reference and immutable reference to the same object exist at the same time, it makes it impossible to set the timestamp while `fetch_cdc` is running.

There is a way out of this issue - you can use a Mutex to protect the lower_timestamp:

```rust
use tokio::sync::Mutex;

pub struct Reader {
    // ...
    upper_timestamp: tokio::sync::Mutex<Option<chrono::Duration>>,
    // ...
}
```

...and `set_lower_timestamp` then may be changed to take an immutable reference - Rust allows multiple immutable references to the same objects to exist:

```rust
pub fn set_lower_timestamp(&mut self, new_lower_timestamp: chrono::Duration) {
    let guard = self.upper_timestamp.lock().await;
    *guard = Some(new_lower_timestamp);
}
```

Adjusting `fetch_cdc` is an exercise for the reader :)

piodul (changes requested):
Some more issues:

- Please don't use unwraps to handle situations which can naturally occur during the execution of the program. For example, network going down may cause session creation or query execution to fail. Those errors should be propagated to the user so that they decide what to do with them.

  Think about unwraps like about asserts - they should always succeed, unless some invariant is broken, or an assumption does not hold (those assumptions should be clearly stated and documented).

- Please fix all the warnings before submitting the code (unless you explicitly say that it is a work in progress). Warnings offer good advice and sometimes catch bugs - for example, in the `fetch_cdc_data` test you forgot to handle the result from `cdc_reader.fetch_cdc` - if you run `cargo check --tests` it will tell you that, among other things.

  Moreover, the CI check won't pass until you fix all the warnings.

scylla-cdc/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+mod reader;
```
Ponewor, line 1: We probably want to make this module public.

Ponewor (changes requested):

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::time;
+use std::time::SystemTime;
+
+use chrono;
+use futures::stream::StreamExt;
+use scylla::frame::value::Timestamp;
+use scylla::{Session, SessionBuilder};
+use tokio::time::sleep;
+
+type StreamID = Vec<u8>;
```
Ponewor, line 12: It would be nice if you used the same type as @macher259 
```rust
use crate::stream_generation::StreamID;
```
Note: it requires @macher259 applying this [suggestion](https://github.com/piodul/scylla-cdc-rust/pull/8/files#r747898988) before.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+                .execute_iter(
+                    query_base.clone(),
+                    (
+                        &self.stream_id_vec,
+                        Timestamp(lower_timestamp),
+                        Timestamp(upper_timestamp),
+                    ),
+                )
+                .await?
+                .into_typed::<(i32, i32, String, String)>();
```
Ponewor, line 104: This cast is incorrect - there are much more columns selected in your query.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+
+            while let Some(row) = rows_stream.next().await {
+                let (pk, t, v, s) = row?;
+                println!("pk: {} t: {} v: {} s: {}", pk, t, v, s);
+            }
```
Ponewor, line 109: Does this even print anything? It certainly does not during your tests.

Ponewor (changes requested):

scylla-cdc/src/reader.rs
piodul: Oh, I just noticed that this function is named `set_lower_timestamp`, but it sets the _upper_ timestamp. The function's name should be changed to `set_upper_timestamp`.

Ponewor: ^Yeah, this problem was already spotted ;)

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    async fn setup(session: &Session) -> Result<Reader, Box<dyn Error>> {
+        create_test_db(session).await;
+        populate_test_db(session).await;
+        let stream_id = get_cdc_stream_id(session).await?;
+
+        let start_timestamp: chrono::Duration =
+            chrono::Duration::from_std(SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?)?;
+        let sleep_interval: time::Duration = time::Duration::new(SLEEP_INTERVAL, 0);
```
Ponewor, line 158: Use `time::Duration::from_secs`.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    pub fn set_lower_timestamp(&mut self, new_lower_timestamp: chrono::Duration) {
+        self.upper_timestamp = Some(new_lower_timestamp);
+    }
+
+    pub async fn fetch_cdc(
+        &self,
+        keyspace: String,
+        table_name: &str,
```
Ponewor, line 53: Is there a particular reason for distinct types here?

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    async fn create_test_db(session: &Session) -> Result<(), Box<dyn Error>> {
+        let mut create_keyspace_query = Query::new(format!(
+            "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class': 'SimpleStrategy', 'replication_factor': 1}};",
+            TEST_KEYSPACE
+        ));
+        create_keyspace_query.set_consistency(Consistency::All);
+
+        session.query(create_keyspace_query, &[]).await?;
```
macher259, line 183: You should wait for schema agreement as in [source](https://github.com/piodul/scylla-cdc-rust/pull/8#issuecomment-966628527)

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,333 @@
...
+        *guard = Some(new_upper_timestamp);
+    }
+
+    pub async fn fetch_cdc(
+        &mut self,
+        keyspace: String,
+        table_name: String,
+    ) -> Result<(), Box<dyn Error>> {
+        let query = format!(
+            "SELECT pk,t,v,s FROM {}.{}_scylla_cdc_log \
```
piodul, line 86: Here, you should use `SELECT *` to select all columns.

scylla-cdc/src/reader.rs
piodul: You shouldn't perform any casts here. The stream will return `Row`s which are untyped wrappers around `Vec<CqlValue>` - please pass the rows directly to the callback (with which you should replace printing the rows on line 124).

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,333 @@
...
+                .into_typed::<(i32, i32, String, String)>();
+
+            while let Some(row) = rows_stream.next().await {
+                let (pk, t, v, s) = row?;
+                println!("pk: {} t: {} v: {} s: {}", pk, t, v, s);
+            }
+
+            if lower_timestamp == upper_timestamp {
+                break;
+            }
```
piodul, line 129: I think you should compare with `self.upper_timestamp` here, not the local variable `upper_timestamp`. Moreover, for safety, the comparison should use `>=` instead of the equality operator.

scylla-cdc/src/reader.rs
piodul: You can look at the tests in `stream_generations.rs` for inspiration on how to initialize the database for the tests: https://github.com/piodul/scylla-cdc-rust/blob/1974fbf3b1ab0b17048835674424560e1af49fff/scylla-cdc/src/stream_generations.rs#L301-L331

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,333 @@
...
+    //     delete_test_data_in_db(&shared_session).await?;
+    //     Ok(())
+    // }
+
+    async fn test_fetch_cdc(
+        shared_session: &Arc<Session>,
+        cdc_reader: &mut StreamReader,
+    ) -> Result<(), Box<dyn Error>> {
+        match cdc_reader
+            .fetch_cdc(TEST_KEYSPACE.parse()?, TEST_TABLE.parse()?)
```
piodul, line 322: Why use `parse()` here?? I think you just want to convert `&str` to `String`? If so, then please just use `to_string()`.

piodul (changes requested):
Leaving another batch of comments.

The PR #8 was merged, please rebase in order to resolve conflicts.

scylla-cdc/src/reader.rs
Gor027: As I understand, we need either the preset `upper_timestamp` or the "safe" `upper_timestamp` calculated at the moment, could you please clarify why `self.upper_timestamp` should be in the comparison.

piodul: The local variables `lower_timestamp` and `upper_timestamp` mean the range of the time window you are reading from in this iteration. The `self.upper_timestamp` means the timestamp at which we want the stream reader to stop. When the stream reader detects that it read all changes up to `self.upper_timestamp`, we want the `fetch_cdc` function to stop.

By the way, I'd suggest renaming `lower_timestamp` -> `window_begin` and `upper_timestamp` (the local variable) -> `window_end` in order to reduce confusion.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        Ok(rows.next().await.unwrap()?)
+    }
+
+    async fn delete_test_data_in_db(session: &Session) -> Result<(), Box<dyn Error>> {
+        session
+            .query(format!("DROP KEYSPACE IF EXISTS {}", TEST_KEYSPACE), ())
+            .await?;
+        Ok(())
+    }
```
piodul, line 258: Please remove this function. It's OK to leave the database in a messy state after a test, and it's hard to ensure that this function is always run at the end of the test (besides, it can fail). It's better to make sure that the environment is cleaned up at the beginning of the test - I can see that the `create_test_db` function already does that.

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                let guard = self.upper_timestamp.lock().await;
+                upper_timestamp = guard.unwrap_or(max(
+                    lower_timestamp,
+                    min(
+                        lower_timestamp + self.window_size,
+                        now - self.safety_interval,
+                    ),
+                ));
+            }
```
piodul, line 85: You can do it in a slightly simpler way:

```rust
let upper_timestamp = {
    let guard = self.upper_timestamp.lock().await;
    guard.unwrap_or(max(
        lower_timestamp,
        min(
            lower_timestamp + self.window_size,
            now - self.safety_interval,
        ),
    ));
}
```

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    async fn delete_test_data_in_db(session: &Session) -> Result<(), Box<dyn Error>> {
+        session
+            .query(format!("DROP KEYSPACE IF EXISTS {}", TEST_KEYSPACE), ())
+            .await?;
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn fetch_cdc_data() -> Result<(), Box<dyn Error>> {
```
piodul, line 261: I don't understand what the tests `fetch_cdc_data` and `fetch_with_set_upper_timestamp` are supposed to check. Could you explain?

scylla-cdc/src/reader.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub async fn set_upper_timestamp(&mut self, new_upper_timestamp: chrono::Duration) {
+        let mut guard = self.upper_timestamp.lock().await;
+        *guard = Some(new_upper_timestamp);
+    }
+
+    pub async fn fetch_cdc<Fut>(
+        &mut self,
+        keyspace: String,
+        table_name: String,
+        after_fetch_callback: fn(RowIterator) -> Fut,
```
piodul, line 55: I don't think a raw function pointer will be enough for tests. `fn(RowIterator) -> Fut` is a raw function type - which means that it doesn't allow capturing state, just like function pointers in C. In tests, you will probably want to record the data which was provided to the callback and perform some assertions on that data (e.g. that the stream reader managed to read a particular amount of rows). In order to do that, the test function and the closure provided by the test will have to share some variables - for example, the callback will append to a Vec, and after `fetch_cdc` finishes, the test function will analyze the data.

Additionally, I'd suggest changing the `fetch_cdc` so that it iterates over the stream itself and passes rows to the callback.

piodul (changes requested):
This PR definitely needs a more comprehensive set of tests. Please add tests which check the following:

1. Insert some data into one stream and test that reading data from a single stream works
1. Insert some data into two streams and that reading data from multiple streams works
1. ~Insert two rows into a stream, set the upper timestamp to a value between the timestamps of those two rows and then check that the reader reads only the first row~
	- ~Hint: you can set a timestamp on a Query or PreparedStatement using the set_timestamp method - Scylla will use this timestamp for the `cdc$timestamp` column.~

EDIT: I think we can skip the last test, I don't think the logic should be as strict.

All those tests should collect received rows into a Vec and assert that the rows are equal to some expected rows.

----

By the way - in the future, please do not use merges to resolve conflicts - rebase your changes onto the newest main instead. Because a merge commit is included in the pull request now, GitHub displays a combination of your commits and some commits from the main branch and it's hard to tell which changes are yours and which are coming from the main branch.

piodul (changes requested):

piodul (commented):
Looks mostly good. There are two small things to fix, then make sure that the CI passes and - after that - I think we can merge.

- The tests you wrote take a lot of time to complete - I measured about 30 seconds. Other developers will want to run those tests several times during development of other features, so it's worthwile to make them run faster in order to reduce the time wasted. I think you can tweak some parameters to make it run faster - I did it on my machine and cut the testing time to about 3 seconds:

	- Reduce `SLEEP_INTERVAL` to 100ms - we are only polling from a small number of streams so polling 10 times per second is perfectly fine
	- Reduce `SAFETY_INTERVAL` to 100ms - you are inserting everything up front with deliberately chosen timestamps, so there is no risk of skipping something accidentally
	- Reduce `to_set_upper_timestamp` to 1 second in places where it was set to 2 seconds - one second should be sufficient.

- Please clean up the commit history and rebase on top of the `main` branch. Ideally, all the changes in the PR should be squashed into two commits - one which moves some common types from `stream_generations.rs` to the new file `cdc_types.rs` - but squashing everything together into one commit is fine, too. You can refer to the discussion on team's Slack channel for tips on how to do it.

.gitignore
```diff
@@ -2,4 +2,4 @@
 /scylla-cdc/target
 /target
 Cargo.lock
-.idea/
+.idea/
```
Ponewor, line 5: Do not remove newline here.

scylla-cdc/src/cdc_types.rs
```diff
@@ -0,0 +1,42 @@
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError};
```
Ponewor, line 1: About the commit message: "Add cdc reader component by provided stream ids" what does it even mean?

Ponewor (changes requested):

scylla-cdc/src/cdc_types.rs
Gor027: I have changed the commit message to include more details, however, if you have other suggestions feel free to tell them.

piodul (approved):

scylla-cdc/src/stream_generations.rs
```diff
@@ -240,10 +208,11 @@ async fn new_distributed_system_query(stmt: String, session: &Session) -> anyhow
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use scylla::statement::Consistency;
     use scylla::SessionBuilder;
 
+    use super::*;
```
Ponewor, line 214: Why?

Issue #5, opened 2021-10-29
Design and implement an interface for consuming changes from CDC log


piodul:
The library is supposed to read changes from the CDC log and feed it to a user-defined callback. For example, each stream reader (ref: #3) may keep an instance of a user-defined object which implements a `Consumer` trait, and calls a method on that interface for every received row.

PR #10, opened 2021-11-12
Implement simple CDC Log printer

### LINKED ISSUE #4: Implement a simple CDC log printer

Ponewor:
Resolves #4 

This is WIP.

I have two questions:

Should Readers for distinct streams work in separate tokio tasks?

Since the only thing this library does is printing to stdout and it's probably not gonna do it anymore in the final version, do we require tests for it at this point? I mean, I could capture stdout somehow, there are surely some ways to achieve it in Rust but is it worth the effort? I've assumed not, so no tests from me this time.

### DIFF
--- a/scylla-cdc-printer/src/printer.rs
+++ b/scylla-cdc-printer/src/printer.rs
@@ -1,0 +1,317 @@
+use std::sync::Arc;
+
+use anyhow;
+use chrono::Duration;
+use core::time;
+use futures::future::RemoteHandle;
+use futures::stream::{FuturesUnordered, StreamExt};
+use futures::FutureExt;
+use scylla::frame::response::result::Row;
+use scylla::Session;
+use scylla_cdc::cdc_types::GenerationTimestamp;
+use tokio::task::JoinError;
+
+use scylla_cdc::reader::StreamReader;
+use scylla_cdc::stream_generations::GenerationFetcher;
+
+pub struct CDCLogPrinter {
+    // Tells the worker to stop
+    // Usage of the "watch" channel will make it possible to change the the timestamp later,
+    // for example if somebody loses patience and wants to stop now not later
+    end_timestamp: tokio::sync::watch::Sender<Duration>,
+}
+
+impl CDCLogPrinter {
+    // Creates a printer and runs it immediately
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        session: &Arc<Session>,
+        keyspace: String,
+        table_name: String,
+        start_timestamp: Duration,
+        end_timestamp: Duration,
+        window_size: Duration,
+        safety_interval: Duration,
+        sleep_interval: time::Duration,
+    ) -> (Self, RemoteHandle<Result<anyhow::Result<()>, JoinError>>) {
+        let (end_timestamp_sender, end_timestamp_receiver) =
+            tokio::sync::watch::channel(end_timestamp);
+        let mut worker = CDCLogPrinterWorker::new(
+            session,
+            keyspace,
+            table_name,
+            start_timestamp,
+            end_timestamp,
+            window_size,
+            safety_interval,
+            sleep_interval,
+            end_timestamp_receiver,
+        );
+        let (_, handle) = tokio::task::spawn(async move { worker.run().await }).remote_handle();
+        let printer = CDCLogPrinter {
+            end_timestamp: end_timestamp_sender,
+        };
+        (printer, handle)
+    }
+
+    // Tell the worker to set the end timestamp and then stop
+    pub fn stop_at(&mut self, when: Duration) {
+        let _ = self.end_timestamp.send(when);
+    }
+
+    // Tell the worker to stop immediately
+    pub fn stop(&mut self) {
+        self.stop_at(Duration::min_value());
+    }
+}
+
+struct CDCLogPrinterWorker {
+    session: Arc<Session>,
+    keyspace: String,
+    table_name: String,
+    start_timestamp: Duration,
+    end_timestamp: Duration,
+    sleep_interval: time::Duration,
+    window_size: Duration,
+    safety_interval: Duration,
+    readers: Vec<Arc<StreamReader>>,
+    end_timestamp_receiver: tokio::sync::watch::Receiver<Duration>,
+}
+
+impl CDCLogPrinterWorker {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        session: &Arc<Session>,
+        keyspace: String,
+        table_name: String,
+        start_timestamp: Duration,
+        end_timestamp: Duration,
+        window_size: Duration,
+        safety_interval: Duration,
+        sleep_interval: time::Duration,
+        end_timestamp_receiver: tokio::sync::watch::Receiver<Duration>,
+    ) -> CDCLogPrinterWorker {
+        CDCLogPrinterWorker {
+            session: Arc::clone(session),
+            keyspace,
+            table_name,
+            start_timestamp,
+            end_timestamp,
+            window_size,
+            safety_interval,
+            sleep_interval,
+            readers: vec![],
+            end_timestamp_receiver,
+        }
+    }
+
+    pub async fn run(&mut self) -> anyhow::Result<()> {
+        let fetcher = Arc::new(GenerationFetcher::new(&self.session));
+        let (mut generation_receiver, _future_handle) = fetcher
+            .clone()
+            .fetch_generations_continuously(self.start_timestamp, self.sleep_interval)
+            .await?;
+
+        fn print_row(row: Row) {
+            println!("{:?}", row.columns);
+        }
+
+        let mut stream_reader_tasks = FuturesUnordered::new();
+
+        let mut next_generation: Option<GenerationTimestamp> = None;
+        let mut had_first_generation: bool = false;
+        let mut err: Option<anyhow::Error> = None;
+
+        loop {
+            tokio::select! {
+                evt = stream_reader_tasks.select_next_some() => {
+                    match evt {
+                        Err(error) => {
+                            err = Some(anyhow::Error::new(error));
+                            self.set_upper_timestamp(chrono::Duration::min_value()).await;
+                        }
+                        Ok(Err(error)) => {
+                            err = Some(error);
+                            self.set_upper_timestamp(chrono::Duration::min_value()).await;
+                        },
+                        _ => {}
+                    }
+                }
+                Some(generation) = generation_receiver.recv(), if next_generation.is_none() => {
+                    next_generation = Some(generation.clone());
+                    had_first_generation = true;
+                    self.set_upper_timestamp(generation.timestamp).await;
+                }
+                Ok(_) = self.end_timestamp_receiver.changed() => {
+                    let timestamp = *self.end_timestamp_receiver.borrow_and_update();
+                    self.end_timestamp = timestamp;
+                    self.set_upper_timestamp(timestamp).await;
+                }
+            }
+
+            if stream_reader_tasks.is_empty() {
+                if err.is_some() {
+                    return Err(err.unwrap());
+                }
+
+                if next_generation.is_some() {
+                    let generation = next_generation.unwrap();
+                    next_generation = None;
+
+                    if generation.timestamp > self.end_timestamp {
+                        return Ok(());
+                    }
+
+                    self.readers = fetcher
+                        .fetch_stream_ids(&generation)
+                        .await?
+                        .iter()
+                        .flatten()
+                        .map(|stream_id| {
+                            Arc::new(StreamReader::new(
+                                &self.session,
+                                vec![stream_id.clone()],
+                                self.start_timestamp,
+                                self.window_size,
+                                self.safety_interval,
+                                self.sleep_interval,
+                            ))
+                        })
+                        .collect();
+
+                    self.set_upper_timestamp(generation.timestamp).await;
+
+                    stream_reader_tasks = self
+                        .readers
+                        .iter()
+                        .map(|reader| {
+                            let reader = Arc::clone(reader);
+                            let keyspace = self.keyspace.clone();
+                            let table_name = self.table_name.clone();
+                            tokio::spawn(async move {
+                                reader.fetch_cdc(keyspace, table_name, print_row).await
+                            })
+                        })
+                        .collect();
+                } else if had_first_generation {
+                    return Ok(());
+                }
+            }
+        }
+    }
+
+    async fn set_upper_timestamp(&self, new_upper_timestamp: Duration) {
+        for reader in self.readers.iter() {
+            reader.set_upper_timestamp(new_upper_timestamp).await;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::time::SystemTime;
+
+    use scylla::batch::Consistency;
+    use scylla::query::Query;
+    use scylla::SessionBuilder;
+    use tokio::time::sleep;
+
+    use super::*;
+
+    const SECOND_IN_MILLIS: i64 = 1_000;
+    const TEST_KEYSPACE: &str = "test";
+    const TEST_TABLE: &str = "t";
+    const SLEEP_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
+    const WINDOW_SIZE: i64 = SECOND_IN_MILLIS / 10 * 3;
+    const SAFETY_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
+
+    fn get_create_table_query() -> String {
+        format!("CREATE TABLE IF NOT EXISTS {}.{} (pk int, t int, v text, s text, PRIMARY KEY (pk, t)) WITH cdc = {{'enabled':true}};",
+                TEST_KEYSPACE,
+                TEST_TABLE
+        )
+    }
+
+    async fn create_test_db(session: &Arc<Session>) -> anyhow::Result<()> {
+        let mut create_keyspace_query = Query::new(format!(
+            "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class': 'SimpleStrategy', 'replication_factor': 1}};",
+            TEST_KEYSPACE
+        ));
+        create_keyspace_query.set_consistency(Consistency::All);
+
+        session.query(create_keyspace_query, &[]).await?;
+        session.await_schema_agreement().await?;
+
+        // Create test table
+        let create_table_query = get_create_table_query();
+        session.query(create_table_query, &[]).await?;
+        session.await_schema_agreement().await?;
+
+        let table_name_under_keyspace = format!("{}.{}", TEST_KEYSPACE, TEST_TABLE);
+        session
+            .query(format!("TRUNCATE {};", table_name_under_keyspace), &[])
+            .await?;
+        session
+            .query(
+                format!("TRUNCATE {}_scylla_cdc_log;", table_name_under_keyspace),
+                &[],
+            )
+            .await?;
+        Ok(())
+    }
+
+    async fn populate_db_with_pk(session: &Arc<Session>, pk: u32) -> anyhow::Result<()> {
+        let table_name_under_keyspace = format!("{}.{}", TEST_KEYSPACE, TEST_TABLE);
+        for i in 0..3 {
+            session
+                .query(
+                    format!(
+                        "INSERT INTO {} (pk, t, v, s) VALUES ({}, {}, 'val{}', 'static{}');",
+                        table_name_under_keyspace, pk, i, i, i
+                    ),
+                    &[],
+                )
+                .await?;
+        }
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_cdc_log_printer() {
+        let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+        let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+        let shared_session = Arc::new(session);
+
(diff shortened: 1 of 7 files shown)


piodul:
> Should Readers for distinct streams work in separate tokio tasks?

Yes, definitely. Running everything on the same task prevents us from using multiple cores.

> Since the only thing this library does is printing to stdout and it's probably not gonna do it anymore in the final version, do we require tests for it at this point? I mean, I could capture stdout somehow, there are surely some ways to achieve it in Rust but is it worth the effort? I've assumed not, so no tests from me this time.

I agree that capturing stdout is a bad idea. However, we would definitely like some tests before this is merged. Perhaps instead of hardcoding printing logic in the stream reader we should add a possibility to provide a custom callback to the reader. The callback interface can be very simple, not as complicated as described in #5, just enough so that it is sufficient to write the tests and be able to print out the changes. Please communicate with @Gor027 and agree upon an interface.

scylla-cdc-printer/src/main.rs
```diff
@@ -1,11 +1,73 @@
...
+    }
+
+    async fn run(&self, keyspace: String, table_name: String, start_timestamp: chrono::Duration) -> Result<(), Box<dyn Error>> {
+        let fetcher = GenerationFetcher::new(&self.session);
+        let mut generation = fetcher.fetch_generation_by_timestamp(&start_timestamp).await?.unwrap_or(loop {
+            match fetcher.fetch_all_generations().await?.last() {
+                None => { sleep(self.sleep_interval).await; }
+                Some(generation) => break generation.clone()
+            }
+        });
```
piodul, line 41: The expression passed to `unwrap_or` will always be evaluated. This means that `fetch_all_generations` will always be called, sometimes unnecessarily.

I suggest moving the logic responsible for getting the first generation to a helper function.

scylla-cdc-printer/src/main.rs
```diff
@@ -1,11 +1,73 @@
...
+        loop {
+            let stream_ids = fetcher.fetch_stream_ids(&generation).await?;
+            for stream_id in stream_ids {
+                let reader = Reader::new(vec![stream_id.clone()], start_timestamp, self.window_size, self.safety_interval, self.sleep_interval);
+                reader.fetch_cdc(keyspace.clone(), table_name.clone()).await?;
+            }
```
piodul, line 47: This doesn't work well. Without the end timestamp set, `fetch_cdc` will keep reading its stream indefinitely. Moreover, what you are doing here is: run `fetch_cdc` for one stream, then after it finishes run it for another stream, etc. There is no concurrency here.

The big idea is that we want to run each reader in a task, independently from each other. This will allow them to utilize multiple cores. In the meantime, your main function monitors if there is a new generation on the horizon - if so, then you should set the end timestamp for all readers, wait until they finish reading and then proceed to the next generation.

Note that being able to concurrently run `fetch_cdc` in one task and calling `set_upper_timestamp` requires this suggestion to be applied to the PR with stream reader: https://github.com/piodul/scylla-cdc-rust/pull/9#discussion_r746949292

piodul (changes requested):

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,246 @@
...
+                    self.start_timestamp,
+                    self.window_size,
+                    self.safety_interval,
+                    self.sleep_interval,
+                );
+                self.readers.push(Arc::new(reader));
+            }
+
+            let handles = self.readers.iter().map(|reader| {
+                tokio::spawn(async move {
```
piodul, line 74: You are using an `async move` block here - because it is marked as `move`, it moves inside itself _all_ variables from the outer scope which are used in the block. Currently, there is going to be a problem with every one of them:

- `reader` - it has type `&Arc<StreamReader>` and is a reference to one of the readers in the `CDCLogPrinter`. Because of it, the async block needs to be destroyed before the Arc pointer to by the reference, otherwise there will be a dangling reference. The futures running in tokio tasks are required to have static lifetimes and rustc will complain because the captured reference will make the future (async block) non-static.
- `keyspace`. `table_name`, `after_fetch_callback` - the closure at line 73 captures those variables from outer scope by reference, but the async block tries to move them in. It won't work.

The solution is to clone those variables _before_ moving them to the async block:

```rust
let handles = self.readers.iter().map(|reader| {
    // The next line works provided that the fetch callback is cloneable
    // This might or might not be something that we want
    let cb = after_fetch_callback.clone();
    let keyspace = keyspace.clone();
    let table_name = table_name.clone();
    tokio::spawn(async move {
        reader
            .fetch_cdc(keyspace.clone(), table_name.clone(), cb)
            .await
    })
});
```

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,246 @@
...
+            sleep_interval,
+            readers: vec![],
+        }
+    }
+
+    pub async fn run(
+        &mut self,
+        keyspace: String,
+        table_name: String,
+        mut after_fetch_callback: impl FnMut(Row),
```
piodul, line 49: Working with a single `after_fetch_callback` is not ideal because it needs exclusive access in order to be called - you would have to put it into an `Arc<Mutex<CallbackType>>` and acquire a mutex before passing each row to it which is inefficient.

I'd recommend using the `ConsumerFactory` and `Consumer` framework we already have - the printer would take a factory and create separate consumers for each of the tokio tasks.

(Besides, a printer taking an arbitrary callback does not look like a printer to me - more like a minimal version of the library)

piodul (changes requested):
There are some issues related to error handling and cancelling tasks:

1. Tokio tasks may leak if you drop a `JoinHandle` before making sure that the task will stop and you have waited for it.

   In case of some tasks (such as the generation fetcher task, if you refactor to use a task for that purpose) you can safely use `remote_handle` to obtain a guard object which will cancel the future running in the task. This means that the future will no longer will be polled, i.e. if the future was preempted at an `.await` point, it may happen that it won't continue and will be destroyed.

   For some tasks this approach is OK - for example, the generation fetcher task - but for the tasks running the user logic _this is not acceptable_ as users will have to account for the possibility of `.await` and this can make it hard to write the error handling logic correctly - what if the consumer needs to perform an async operation in error handling?

   In case of those tasks we should stop them gracefully. We can adapt a policy that `consume_cdc` will not be interrupted in any way - the `StreamReader` should detect that it was stopped and shut down between `consume_cdc` invocations. For the time being, you can implement it hackily by setting the upper_timestamp to the smallest timestamp and we can implement something more proper in a followup.

2. Your code needs to react to errors from tasks even if you are currently waiting for something else, e.g. for a new generation. You need to wait for two things in parallel: the next generation and until StreamReaders return (possibly with an error). You can achieve it by using the [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) macro. If you need some examples on how to use it with `FuturesUnordered` and `tokio::sync::mpsc`, we have some `tokio::select!` invocations in the Rust driver code, you can look there.

scylla-cdc-printer/src/printer.rs
piodul: You can write a test factory which takes a callback and returns consumers that use the callback internally. The consumers would share the callback via `Arc<Mutex<Callback>>`.

Here is a snippet of code to make it more clear what I mean:

```rust
struct TestConsumerFactory {
    cb: Arc<Mutex<dyn FnMut(CDCRow)>>,
}

impl ConsumerFactory for TestConsumerFactory {
    fn new_consumer(&self) -> Box<dyn Consumer> {
        Box::new(TestConsumer { cb: Arc::clone(&self.cb) }
    }
}

struct TestConsumer {
    cb: Arc<Mutex<dyn FnMut(CDCRow)>>,
}

#[async_trait]
impl Consumer for TestConsumer {
    async fn consume_cdc(&mut self, data: CDCRow) {
        (self.cb.lock().unwrap())(data);
    }
}
```

Although perhaps using the factory + consumers now may be an overkill for the printer and its tests, they will be required for the minimal version of the library, so if you do it now you will have less work later.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,246 @@
...
+                            break GenerationTimestamp { timestamp: self.end_timestamp };
+                        }
+                        sleep(self.sleep_interval).await
+                    }
+                }
+            };
+
+            self.set_upper_timestamp(generation.timestamp).await;
+
+            futures::future::try_join_all(handles).await?;
```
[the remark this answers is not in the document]
Ponewor: What is `consume_cdc` you mention?

How the errors should be handled?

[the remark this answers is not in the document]
piodul: `consume_cdc` is a method of the `Consumer` trait. You should create a separate consumer for each tokio task and feed it CDC rows using the `consume_cdc` method.

Ideally, in case of an error the whole library should stop (gracefully) and return the error.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,273 @@
...
+        let end_timestamp = self.end_timestamp;
+        tokio::spawn(async move {
+            generation = loop {
+                match fetcher.fetch_next_generation(&generation).await {
+                    Ok(Some(generation)) => break generation,
+                    Ok(None) => {
+                        let now = SystemTime::now()
+                            .duration_since(SystemTime::UNIX_EPOCH)
+                            .unwrap();
+                        if now + sleep_interval >= end_timestamp.to_std().unwrap() {
```
piodul, line 96: I don't think the generation fetcher task to know about the end timestamp. It should continuously push generations to the `generation_sender`. When `CDCLogPrinter::run` no longer needs it, it should drop the RemoteHandle corresponding to the task and the task will stop itself.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,273 @@
...
+            let reader = StreamReader::new(
+                &self.session,
+                vec![stream_id.clone()],
+                self.start_timestamp,
+                self.window_size,
+                self.safety_interval,
+                self.sleep_interval,
+            );
+            self.readers.push(Arc::new(reader));
+        }
```
piodul, line 67: New readers should be created every time a new generation starts. Each generation has different stream IDs.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,273 @@
...
+                        }
+                        _ => {}
+                    }
+                }
+                generation = generation_receiver.recv() => {
+                    self.set_upper_timestamp(generation.unwrap().timestamp).await;
+                }
+
+            }
+
```
piodul, line 127: Here, you should wait until all `futures` are finished. We must wait until the streams are consumed to the end before we start reading next generation.

Next, if there was an error returned from any of the tasks, you should return it, too. Otherwise, you should create new stream readers and go back to the start of the loop.

piodul (changes requested):

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,273 @@
...
+                let table_name = table_name.clone();
+                tokio::spawn(async move { reader.fetch_cdc(keyspace, table_name, print_row).await })
+            })
+            .collect();
+
+        let (generation_sender, mut generation_receiver) = mpsc::channel(1);
+
+        let sleep_interval = self.sleep_interval;
+        let end_timestamp = self.end_timestamp;
+        tokio::spawn(async move {
```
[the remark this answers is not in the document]
piodul: Suggestion: it might also make sense (or not) to move the `get_first_generation` logic to the beginning of the task so that the first generation is pushed to the generations channel, too, and can be handled in a unified way.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,273 @@
...
+                self.start_timestamp,
+                self.window_size,
+                self.safety_interval,
+                self.sleep_interval,
+            );
+            self.readers.push(Arc::new(reader));
+        }
+
+        fn print_row(row: Row) {
+            println!("{}", row.columns.len());
```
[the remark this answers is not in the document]
Ponewor: I have completely forgotten about `:?` :flushed:

scylla-cdc-printer/src/printer.rs
Ponewor: Does it matter if we wait for the futures to finish right here where you have mentioned it or just after we realise there exists a new generation, or is it simply a stylistic choice?

piodul: What I meant is that you need to wait for all `futures` to finish after you `set_upper_timestamp` for all stream readers. AFAIK we have two cases here: either we realized there is a new generation or we observed that one of the `futures` returned an error.

I suggested placing the waiting code here because it looked like it would cover both of those cases. However, if you are rearranging the logic then it may make sense to place it somewhere else.

Ponewor:
What is wrong with clippy? It does not complain locally...

piodul:
> What is wrong with clippy? It does not complain locally...

Most likely you have an old version of clippy installed. If you installed Rust through rustup, then `rustup update` should update your Rust installation along with clippy.

piodul:
Hmm, now that I looked more closely in the CI logs it looks like clippy complains about code unrelated to this PR... Most likely the clippy version used in our CI got updated recently. Please fix the complaints and send in a separate PR.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,294 @@
...
+            end_timestamp,
+            window_size,
+            safety_interval,
+            sleep_interval,
+            readers: vec![],
+            end_timestamp_receiver,
+        }
+    }
+
+    pub async fn run(&mut self) -> anyhow::Result<()> {
```
piodul, line 108: This version of `CDCLogPrinterWorker::run` is a big step in a good direction. The event loop pattern is a good idea, but here are still some issues related to how generation change is handled and when the function is supposed to exit. Nevertheless, I think we are close to the finish line.

To help you out, here is how I think a correct loop should look like. Given this additional state:

```rust
let next_generation: Option<Generation> = None;
let had_first_generation: bool = false;
let err: Option<anyhow::Error> = None;
```

...the loop iteration should look like this:

- Wait for an event
	- When a task has finished:
		- If there was an error, store this error in `err` for later
	- When a new generation appears (wait for this event only if `next_generation` is empty - you can put an `if` in the `select!` macro, see the [docs](https://docs.rs/tokio/latest/tokio/macro.select.html)):
		- Store the generation in `next_generation` 
		- Set `had_first_generation` to true
		- Update the upper timestamps of `self.readers`
	- When the end timestamp is updated:
		- Set the end timestamp field
		- Update the upper timestamps of `self.readers`

- If there are no tasks running:
	- If there is `err`:
		- Return the error
	- If there is `next_generation`:
		- If the `next_generation` is larger than the end timestamp:
			- Return success
		- Else:
			- Spawn new tasks
			- Clear `next_generation`
	- Else if `had_first_generation`:
		- Return success

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,294 @@
...
+                .query(
+                    format!(
+                        "INSERT INTO {} (pk, t, v, s) VALUES ({}, {}, 'val{}', 'static{}');",
+                        table_name_under_keyspace, pk, i, i, i
+                    ),
+                    &[],
+                )
+                .await?;
+        }
+        session.await_schema_agreement().await?;
```
piodul, line 251: Waiting for schema agreement is only necessary after you make schema changes (create/modify a keyspace/table). You aren't doing any schema changes in this function.

piodul (changes requested):
The overall shape looks good, however there are still some changes to be made in `CDCLogPrinterWorker::run`. I left a detailed comment which describes how it should be modified.

As for the tests... I think we can skip automated tests for now and postpone them for https://github.com/piodul/scylla-cdc-rust/issues/12. For this PR I'd like you to at least perform some manual checks and observe if printed messages make sense. The following scenarios should be sufficient:

1. Start the printer, insert 1000 keys, add a node, insert 1000 more keys after the generation changes. Verify that the printer processed the first 1000 keys before the latter 1000 keys.
1. Run the printer again on the same cluster, starting from the same timestamp as in the previous test. Verify the same assertion as in the previous test.
1. Run the printer again on the same cluster but give it an end_timestamp which is before the second generation starts. Verify that only the first 1000 keys are printed.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,294 @@
...
+                    })
+                    .collect();
+
+                    futures = self.readers
+                        .iter()
+                        .map(|reader| {
+                            let reader = Arc::clone(reader);
+                            let keyspace = self.keyspace.clone();
+                            let table_name = self.table_name.clone();
+                            tokio::spawn(async move { reader.fetch_cdc(keyspace, table_name, print_row).await})
```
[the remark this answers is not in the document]
Ponewor: I am surprised neither `rustfmt` nor `clippy` caught it :(

scylla-cdc-printer/src/printer.rs
Ponewor: Do we still want to stop all the tasks if one of them has failed (`set_upper_timestamp(chrono::Duration::min_value())`) or do we abandon this idea?

Ponewor: Thank you for detailed break down. 

Why are we returning when there is no `next_generation` and `had_first_generation` is `true`? What makes us sure that there is no `next_generation` coming in the next iteration? I do not get it. 

More generally: why are we handling `first_generation` differently?

scylla-cdc/src/stream_generations.rs
```diff
@@ -172,6 +178,42 @@ impl GenerationFetcher {
...
+        let generation = self
+            .fetch_generation_by_timestamp(&start_timestamp)
+            .await?
+            .unwrap_or(loop {
+                match self.fetch_all_generations().await?.last() {
+                    None => sleep(sleep_interval).await,
+                    Some(generation) => break generation.clone(),
+                }
+            });
+        generation_sender.send(generation.clone()).await.unwrap();
```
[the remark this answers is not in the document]
Ponewor: Yeah but it just got considerably messier.

scylla-cdc-printer/src/printer.rs
piodul: > Do we still want to stop all the tasks if one of them has failed (`set_upper_timestamp(chrono::Duration::min_value())`) or do we abandon this idea?

Yes, we do. Sorry, I missed it in the break down - I updated it just now.

piodul: > Thank you for detailed break down.
> 
> Why are we returning when there is no `next_generation` and `had_first_generation` is `true`? What makes us sure that there is no `next_generation` coming in the next iteration? I do not get it.

I see there is a potentially problematic situation:

1. We are reading generation `G_1`,
2. The printer is told to stop at timestamp `T`,
3. A generation `G_2` appears which is less than `T`,
4. Due to some issues fetching generations doesn't succeed for some time and the printer doesn't learn about `G_2`,
5. The printer reaches timestamp `T` in generation `G_1`,
6. The printer decides to stop, despite not reading data from range `G_2..T` in `G_2`.

I talked with @kbr- and he suggested triggering a generation fetch and making sure it succeeds when we reach point (5), so that we know if there are more generations to process. Generations are advertised 1 minute before, so the assumption is that a successful fetch will inform us about the problematic generation.

However, not to block the progress of this PR further, I think we can leave it as a FIXME in the code. Please include a short description of the problem and the solution/link to this GitHub comment.

> 
> More generally: why are we handling `first_generation` differently?

The `had_first_generation` flag was supposed to differentiate two cases: 1) we are waiting for the first generation 2) we reached the end timestamp and there was no new generation. It was supposed to prevent a situation where the printer quits due to the end timestamp being set before the first generation arrives.

Perhaps we can get rid of it if we wait for the first generation separately (despite me telling you otherwise in the previous round of review). Waiting could be aborted in case a minimum timestamp is set as the end timestamp.

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,317 @@
...
+            }
+
+            if stream_reader_tasks.is_empty() {
+                if err.is_some() {
+                    return Err(err.unwrap());
+                }
+
+                if next_generation.is_some() {
+                    let generation = next_generation.unwrap();
+                    next_generation = None;
```
piodul, line 159: This is:

```rust
if let Some(next_generation) = next_generation.take() {
    // ...
```

scylla-cdc-printer/src/printer.rs
```diff
@@ -0,0 +1,317 @@
...
+                    let timestamp = *self.end_timestamp_receiver.borrow_and_update();
+                    self.end_timestamp = timestamp;
+                    self.set_upper_timestamp(timestamp).await;
+                }
+            }
+
+            if stream_reader_tasks.is_empty() {
+                if err.is_some() {
+                    return Err(err.unwrap());
+                }
```
piodul, line 155: Why not:

```rust
if let Some(err) = err {
    return Err(err);
}
```

piodul:
OK, I think it is good enough for merge. Let's do the remaining issues (cosmetic changes + tests) in a followup.

piodul (approved):

PR #13, opened 2021-11-12
Interface for consuming changes

procrastinationfighter:
Fixes: #5 

This pull request adds an interface for consuming changes from CDC table. 
The module `consumer` provides a `Consumer` trait and a `ConsumerFactory` trait that define interface used by the reader component. Data from the table is passed to a consumer in a `CDCRow` object. `CDCSchema` is used to create new `CDCRow` objects.

### DIFF
--- a/scylla-cdc/src/consumer.rs
+++ b/scylla-cdc/src/consumer.rs
@@ -1,0 +1,468 @@
+use num_enum::TryFromPrimitive;
+use scylla::frame::response::result::{ColumnSpec, CqlValue, Row};
+use std::collections::HashMap;
+
+pub trait Consumer {
+    fn consume_cdc(&mut self, data: CDCRow);
+}
+
+pub trait ConsumerFactory {
+    fn new_consumer(&self) -> Box<dyn Consumer>;
+}
+
+#[derive(Debug, Eq, PartialEq, TryFromPrimitive)]
+#[repr(i8)]
+pub enum OperationType {
+    PreImage,
+    RowUpdate,
+    RowInsert,
+    RowDelete,
+    PartitionDelete,
+    RowRangeDelInclLeft,
+    RowRangeDelExclLeft,
+    RowRangeDelInclRight,
+    RowRangeDelExclRight,
+    PostImage,
+}
+
+pub struct CDCRowSchema {
+    // The usize values are indices of given values in the Row.columns vector.
+    pub(crate) stream_id: usize,
+    pub(crate) time: usize,
+    pub(crate) batch_seq_no: usize,
+    pub(crate) end_of_batch: usize,
+    pub(crate) operation: usize,
+    pub(crate) ttl: usize,
+
+    // These HashMaps map names of columns in the observed table to matching columns in the CDC table.
+    // The usize value is an index of the column in an internal vector inside of the CDCRow struct.
+
+    // Maps name of a column to a matching column in the CDC table.
+    pub(crate) mapping: HashMap<String, usize>,
+    // Maps name of a column to a column that tells if value in this column was deleted.
+    pub(crate) deleted_mapping: HashMap<String, usize>,
+    // Maps name of a collection column to a column that tells
+    // if elements from this collection were deleted.
+    pub(crate) deleted_el_mapping: HashMap<String, usize>,
+}
+
+const STREAM_ID_NAME: &str = "cdc$stream_id";
+const TIME_NAME: &str = "cdc$time";
+const BATCH_SEQ_NO_NAME: &str = "cdc$batch_seq_no";
+const END_OF_BATCH_NAME: &str = "cdc$end_of_batch";
+const OPERATION_NAME: &str = "cdc$operation";
+const TTL_NAME: &str = "cdc$ttl";
+const IS_DELETED_PREFIX: &str = "cdc$deleted_";
+const ARE_ELEMENTS_DELETED_PREFIX: &str = "cdc$deleted_elements_";
+
+impl CDCRowSchema {
+    pub fn new(specs: &[ColumnSpec]) -> CDCRowSchema {
+        let mut stream_id = 0;
+        let mut time = 0;
+        let mut batch_seq_no = 0;
+        let mut end_of_batch = 0;
+        let mut operation = 0;
+        let mut ttl = 0;
+        let mut mapping: HashMap<String, usize> = HashMap::new();
+        let mut deleted_mapping: HashMap<String, usize> = HashMap::new();
+        let mut deleted_el_mapping: HashMap<String, usize> = HashMap::new();
+
+        let mut j = 0;
+
+        // Hashmaps will have indices of data in a new vector without the hardcoded values.
+        for (i, spec) in specs.iter().enumerate() {
+            match spec.name.as_str() {
+                // spec.name is public since 0.3.0 driver version, it didn't work on 0.2.1.
+                STREAM_ID_NAME => stream_id = i,
+                TIME_NAME => time = i,
+                BATCH_SEQ_NO_NAME => batch_seq_no = i,
+                END_OF_BATCH_NAME => end_of_batch = i,
+                OPERATION_NAME => operation = i,
+                TTL_NAME => ttl = i,
+                x => {
+                    if let Some(stripped) = x.strip_prefix(ARE_ELEMENTS_DELETED_PREFIX) {
+                        deleted_el_mapping.insert(stripped.to_string(), j);
+                    } else if let Some(stripped) = x.strip_prefix(IS_DELETED_PREFIX) {
+                        deleted_mapping.insert(stripped.to_string(), j);
+                    } else {
+                        mapping.insert(x.to_string(), j);
+                    }
+
+                    j += 1;
+                }
+            }
+        }
+
+        CDCRowSchema {
+            stream_id,
+            time,
+            batch_seq_no,
+            end_of_batch,
+            operation,
+            ttl,
+            mapping,
+            deleted_mapping,
+            deleted_el_mapping,
+        }
+    }
+}
+
+pub struct CDCRow<'schema> {
+    pub stream_id: Vec<u8>,
+    pub time: uuid::Uuid,
+    pub batch_seq_no: i32,
+    pub end_of_batch: bool,
+    pub operation: OperationType,
+    // Can be NULL in the database.
+    pub ttl: Option<i64>,
+    data: Vec<Option<CqlValue>>,
+    // Maps element name to its index in the data vector.
+    schema: &'schema CDCRowSchema,
+}
+
+impl CDCRow<'_> {
+    pub fn from_row(row: Row, schema: &CDCRowSchema) -> CDCRow {
+        // If cdc read was successful, these default values will not be used.
+        let mut stream_id = vec![];
+        let mut time = uuid::Uuid::default();
+        let mut batch_seq_no = i32::MAX;
+        let mut end_of_batch = false;
+        let mut operation = OperationType::PreImage;
+        let mut ttl = None;
+
+        let data_count =
+            schema.mapping.len() + schema.deleted_mapping.len() + schema.deleted_el_mapping.len();
+        let mut data: Vec<Option<CqlValue>> = Vec::with_capacity(data_count);
+
+        for (i, column) in row.columns.into_iter().enumerate() {
+            if i == schema.stream_id {
+                stream_id = column.unwrap().into_blob().unwrap();
+            } else if i == schema.time {
+                time = column.unwrap().as_uuid().unwrap();
+            } else if i == schema.batch_seq_no {
+                batch_seq_no = column.unwrap().as_int().unwrap();
+            } else if i == schema.end_of_batch {
+                end_of_batch = column.unwrap().as_boolean().unwrap()
+            } else if i == schema.operation {
+                operation = OperationType::try_from(column.unwrap().as_tinyint().unwrap()).unwrap();
+            } else if i == schema.ttl {
+                ttl = column.map(|ttl| ttl.as_bigint().unwrap());
+            } else {
+                data.push(column);
+            }
+        }
+
+        CDCRow {
+            stream_id,
+            time,
+            batch_seq_no,
+            end_of_batch,
+            operation,
+            ttl,
+            data,
+            schema,
+        }
+    }
+
+    /// Allows to get a value from the column that corresponds to the logged table.
+    /// Returns None if the value is null.
+    /// Panics if the column does not exist in this table.
+    /// To check if such column exists, use column_exists() method.
+    pub fn get_value(&self, name: &str) -> &Option<CqlValue> {
+        self.schema
+            .mapping
+            .get(name)
+            .map(|id| &self.data[*id])
+            .unwrap()
+    }
+
+    /// Allows to get info if a value was deleted in this operation.
+    /// Panics if the column does not exist in this table
+    /// or the column is a part of primary key (because these values can't be deleted).
+    /// To check if such column exists, use column_deletable() method.
+    pub fn is_value_deleted(&self, name: &str) -> bool {
+        self.schema
+            .deleted_mapping
+            .get(name)
+            .map(|id| self.data[*id].is_some())
+            .unwrap()
+    }
+
+    /// Allows to get deleted elements from a collection.
+    /// Returns empty slice if the value is null.
+    /// Panics if the column does not exist in this table or is not a collection.
+    /// To check if such column exists, use collection_exists() method.
+    pub fn get_deleted_elements(&self, name: &str) -> &[CqlValue] {
+        let val = self
+            .schema
+            .deleted_el_mapping
+            .get(name)
+            .map(|id| self.data[*id].as_ref().map(|val| val.as_set().unwrap()))
+            .unwrap();
+        match val {
+            Some(vec) => vec,
+            None => &[],
+        }
+    }
+
+    pub fn column_exists(&self, name: &str) -> bool {
+        self.schema.mapping.contains_key(name)
+    }
+
+    pub fn column_deletable(&self, name: &str) -> bool {
+        self.schema.deleted_mapping.contains_key(name)
+    }
+
+    pub fn collection_exists(&self, name: &str) -> bool {
+        self.schema.deleted_el_mapping.contains_key(name)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    // Because we are planning to extract a common setup to all tests,
+    // the setup for this module is based on generation fetcher's tests.
+
+    use super::*;
+    use scylla::batch::Consistency;
+    use scylla::query::Query;
+    use scylla::{Session, SessionBuilder};
+    // These tests should be indifferent to things like number of Scylla nodes,
+    // so if run separately, they can be tested on one Scylla instance.
+
+    const TEST_SINGLE_VALUE_TABLE: &str = "ConsumerTest.single_value";
+    const TEST_SINGLE_COLLECTION_TABLE: &str = "ConsumerTest.single_collection";
+    const TEST_KEYSPACE: &str = "ConsumerTest";
+    const CDC_CONFIG: &str = "{'enabled': 'true'}";
+    const TEST_SINGLE_VALUE_CDC_TABLE: &str = "ConsumerTest.single_value_scylla_cdc_log";
+    const TEST_SINGLE_COLLECTION_CDC_TABLE: &str = "ConsumerTest.single_collection_scylla_cdc_log";
+
+    fn construct_single_value_table_query() -> String {
+        format!(
+            "
+    CREATE TABLE IF NOT EXISTS {}(
+    pk int,
+    ck int,
+    v int,
+    PRIMARY KEY(pk, ck)) WITH cdc = {};",
+            TEST_SINGLE_VALUE_TABLE, CDC_CONFIG
+        )
+    }
+
+    async fn populate_single_value_table(session: &Session) {
+        // We want to use ttl, because we check its value in some tests.
+        session
+            .query(
+                format!(
+                    "INSERT INTO {} (pk, ck, v) VALUES ({}, {}, {}) USING TTL {};",
+                    TEST_SINGLE_VALUE_TABLE, 1, 2, 3, 86400
+                ),
+                (),
+            )
+            .await
+            .unwrap();
+    }
+
+    fn construct_single_collection_table_query() -> String {
+        format!(
+            "
+    CREATE TABLE IF NOT EXISTS {}(
+    pk int,
+    ck int,
+    vs set<int>,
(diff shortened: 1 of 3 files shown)


scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub end_of_batch: bool,
+    pub operation: OperationType,
+    // Can be NULL in the database.
+    pub ttl: Option<i64>,
+    data: Vec<CqlValue>,
+    // Maps element name to its index in the data vector.
+    mapping: HashMap<String, usize>,
+}
+
+const STREAM_ID_NAME: &'static str = "cdc$stream_id";
```
macher259, line 34: I think constants have `static` lifetime by default.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+        // [TODO]: Create a way to return a generic value, maybe use an inspiration from Driver's code.
+        let index = self.mapping.get(name);
+        return match index {
+            Some(id) => Some(self.data[id.clone()]),
+            None => None,
+        }
+    }
+
+    // Allows to get info if a value was deleted in this operation.
+    pub fn is_value_deleted(&self, name: &str) -> Option<Option<bool>> {
```
piodul, line 106: What is the meaning of `Option<Option<bool>>` here? I think it should just return `Option<bool>`:

- `None` => there is no such column,
- `Some(false)` => the column was not deleted,
- `Some(true)` => the column was deleted.

Tip: look here: https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#advanced-column-types. If the `cdc$deleted_xyz` column is `null` it means that the corresponding column was _not_ deleted, if it's `true` - then it means that it was.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+        let index = self.mapping.get(&(IS_DELETED_PREFIX.to_owned() + name));
+        return match index {
+            Some(id) => Some(self.data[id.clone()].as_boolean()),
+            None => None,
+        }
+    }
```
piodul, line 112: We need another getter function, for the `cdc$deleted_elements_xyz` function. See the "Advanced column types" section of the CDC documentation for more info: https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#advanced-column-types.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
+use std::collections::HashMap;
+use scylla::frame::response::result::{ColumnSpec, CqlValue, Row};
+
+pub trait Consumer {
+    fn consume_cdc(&self, data: CDCRow);
+}
```
piodul, line 6: Two things:

- We want to allow consumers to keep some internal state an probably modify it. Therefore, `consume_cdc` should take `self` as a mutable reference.
- We want to allow the library to read streams from multiple threads, therefore we won't be sharing a single instance of `Consumer` for processing all streams - instead, each stream (or each group of streams) will get a separate `Consumer` instance. Because the library is responsible for handling stream IDs and not the user, it should have ability to create new `Consumers` itself. The point is that we need to define a `ConsumerFactory`, too.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+            end_of_batch,
+            operation,
+            ttl,
+            data,
+            mapping,
+        }
+    }
+
+    // Allows to get a value from the column that corresponds to the logged table.
+    pub fn get_value<T>(&self, name: &str) -> Option<Option<T>> {
```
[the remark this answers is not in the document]
piodul: Another thing: if you are explaining complex, non-obvious types such as `Option<Option<CqlValue>>`, then you should clearly state in the docstring comment what the result of the function exactly means.

piodul (changes requested):

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct CDCRow {
+    pub stream_id: Vec<u8>,
+    pub time: uuid::Uuid,
+    pub batch_seq_no: i32,
+    pub end_of_batch: bool,
+    pub operation: OperationType,
+    // Can be NULL in the database.
+    pub ttl: Option<i64>,
+    data: Vec<CqlValue>,
```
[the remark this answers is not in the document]
procrastinationfighter: That's fine, but I saw that CqlValue can be an `Empty` - what's the difference?

scylla-cdc/src/customer.rs
procrastinationfighter: So, `cdc$deleted_xyx` is never `false`? I thought that it could be, and this is why I did it in similar way as in `get_value`.

scylla-cdc/src/customer.rs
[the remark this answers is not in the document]
piodul: It's really strange, but for some CQL types "Empty" is an additional, valid value which is different from null - see [here](https://github.com/scylladb/scylla-rust-driver/issues/278#issue-965068630) for a bit more context.

scylla-cdc/src/customer.rs
piodul: Yes, AFAIK nulls don't take up space on disk, so they are used here instead of `false`.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
...
+    PreImage = 0,
+    RowUpdate = 1,
+    RowInsert = 2,
+    RowDelete = 3,
+    PartitionDelete = 4,
+    RowRangeDelInclLeft = 5,
+    RowRangeDelExclLeft = 6,
+    RowRangeDelInclRight = 7,
+    RowRangeDelExclRight = 8,
+    PostImage = 9,
```
Ponewor, line 18: I know this comment is redundant at this point after @piodul's suggestion to use num_enum but for the record:
you don't have to be explicit with integer values here, they are assigned this way by default.

scylla-cdc/src/customer.rs
```diff
@@ -0,0 +1,113 @@
+use std::collections::HashMap;
```
Ponewor, line 1: You probably wanted to name the file consumer.rs, didn't you?

scylla-cdc/src/customer.rs
procrastinationfighter: That's right, I can't decide which option looks better, so I will just leave it for now.

procrastinationfighter: Update: Clippy recommended me to remove it.

procrastinationfighter:
I've applied some changes to the code - it compiles now, however it was not tested yet, so it's still marked as a draft - I will mark this request as ready for review as soon as I add the tests.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+                batch_seq_no = column.unwrap().as_int().unwrap();
+            } else if i == schema.end_of_batch {
+                end_of_batch = column.unwrap().as_boolean().unwrap()
+            } else if i == schema.operation {
+                operation = OperationType::try_from(column.unwrap().as_tinyint().unwrap()).unwrap();
+            } else if i == schema.ttl {
+                ttl = column.map(|ttl| ttl.as_bigint().unwrap());
+            } else {
+                data.push(column);
+            }
```
procrastinationfighter, line 151: I don't know why, but when I tried to do something like that:
`match i {
    schema.stream_id = > (...)
}`
it just didn't work - as if the variable `schema` was out of scope.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+            0,
+            true,
+            OperationType::PreImage,
+            None,
+        );
+        let mut data: Vec<Option<CqlValue>> = vec![];
+        let mut i = 0;
+
+        // .iter().enumerate() can't be used here, because it doesn't take ownership of taken value.
+        for column in row.columns.into_iter() {
```
macher259, line 135: I think you don't have to use `into_iter` here.

macher259, line 135: I think you don't have to use `into_iter` here.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+    mapping: HashMap<String, usize>,
+    deleted_mapping: HashMap<String, usize>,
+    deleted_el_mapping: HashMap<String, usize>,
+}
+
+impl CDCRow {
+    pub fn from_row(row: Row, schema: &CDCRowSchema) -> CDCRow {
+        let mapping = schema.mapping.clone();
+        let deleted_mapping = schema.deleted_mapping.clone();
+        let deleted_el_mapping = schema.deleted_el_mapping.clone();
```
piodul, line 119: It's really sad that construction of a CDCRow requires cloning three hashmaps.

Let's change the CDCRow so that it keeps a reference to a CDCRowSchema:

```rust
pub struct CDCRow<'schema>
    // ...
    schema: &'schema CDCRowSchema,
```

Then, in the getter methods you can refer to the mappings through the `schema` reference.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+        // If cdc read was successful, these default values will not be used.
+        // [TODO]: Test it.
+        let (mut stream_id, mut time, mut batch_seq_no, mut end_of_batch, mut operation, mut ttl) = (
+            vec![],
+            uuid::Uuid::default(),
+            0,
+            true,
+            OperationType::PreImage,
+            None,
+        );
```
piodul, line 130: I think it's more readable if you declare each variable in a different line:

```rust
let mut stream_id = vec![];
let mut time = uuid::Uuid::default();
// ...
```

It will be easier to associate which value corresponds to which variable.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+        let mut data: Vec<Option<CqlValue>> = vec![];
+        let mut i = 0;
+
+        // .iter().enumerate() can't be used here, because it doesn't take ownership of taken value.
```
piodul, line 134: > .iter().enumerate() can't be used here, because it doesn't take ownership of taken value.

How about `.into_iter().enumerate()`?

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+        // [TODO]: Test it.
+        let (mut stream_id, mut time, mut batch_seq_no, mut end_of_batch, mut operation, mut ttl) = (
+            vec![],
+            uuid::Uuid::default(),
+            0,
+            true,
+            OperationType::PreImage,
+            None,
+        );
+        let mut data: Vec<Option<CqlValue>> = vec![];
```
piodul, line 131: The final size of the vector is known up front, so you can preallocate:

```rust
let mut data: Vec<Option<CqlValue>> = Vec::with_capacity(mapping.len());
```

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+                TIME_NAME => time = i,
+                BATCH_SEQ_NO_NAME => batch_seq_no = i,
+                END_OF_BATCH_NAME => end_of_batch = i,
+                OPERATION_NAME => operation = i,
+                TTL_NAME => ttl = i,
+                x => {
+                    if x.starts_with(IS_DELETED_PREFIX) {
+                        deleted_mapping.insert(x.to_string(), j);
+                    } else if x.starts_with(ARE_ELEMENTS_DELETED_PREFIX) {
+                        deleted_el_mapping.insert(x.to_string(), j);
```
piodul, line 77: Strip the prefix here, too.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+        //          The driver does not provide the vector of ColumnSpec for every query yet,
+        //          and therefore we can't map the result
+        //          to something other than a hardcoded tuple.
+
+        let mut mapping: HashMap<String, usize> = HashMap::new();
+        let mut deleted_mapping: HashMap<String, usize> = HashMap::new();
+        let mut deleted_el_mapping: HashMap<String, usize> = HashMap::new();
+        let (mut stream_id, mut time, mut batch_seq_no, mut end_of_batch, mut operation, mut ttl):
+            (usize, usize, usize, usize, usize, usize) = (0, 0, 0, 0, 0, 0);
+        let mut j = 0;
```
piodul, line 61: Please initialize each variable in a separate line. It will be more readable this way.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,194 @@
...
+                OPERATION_NAME => operation = i,
+                TTL_NAME => ttl = i,
+                x => {
+                    if x.starts_with(IS_DELETED_PREFIX) {
+                        deleted_mapping.insert(x.to_string(), j);
+                    } else if x.starts_with(ARE_ELEMENTS_DELETED_PREFIX) {
+                        deleted_el_mapping.insert(x.to_string(), j);
+                    } else {
+                        mapping.insert(x.to_string(), j);
+                    }
```
piodul, line 80: By the way, you can write this logic without using any unwraps:

```rust
if let Some(stripped) = x.strip_prefix(ARE_ELEMENTS_DELETED_PREFIX) {
    deleted_el_mapping.insert(stripped.to_string(), j);
} else if let Some(stripped) = x.strip_prefix(IS_DELETED_PREFIX) {
    deleted_mapping.insert(stripped.to_string(), j);
} else {
    mapping.insert(x.to_string(), j);
}
```

piodul: Additionally, please note that IS_DELETED_PREFIX itself is a prefix of ARE_ELEMENTS_DELETED_PREFIX - when  analyzing the variable name, you need to first check for the ARE_ELEMENTS_DELETED_PREFIX and only then for IS_DELETED_PREFIX.

piodul (changes requested):

procrastinationfighter:
I applied suggested fixes and created some tests for the module. I also have a writeup for a test that checks if `CDCRow` is created properly (without connecting to the database, with prepared `Row` and `CDCRowSchema`), but I didn't finish it, because I changed my opinion about its usefulness and writing it was tiring. If you think that such a test is a good idea, I will finish it.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,474 @@
...
+            .deleted_mapping
+            .get(name)
+            .map(|id| self.data[*id].is_some())
+    }
+
+    /// Allows to get deleted elements from a collection.
+    /// Returns None if there is no collection column with such name.
+    /// Returns Some(None) if there is such collection, but nothing was deleted from it.
+    /// Otherwise returns Some(Some(x)) where x is a reference to vector containing deleted values.
+    pub fn get_deleted_elements(&self, name: &str) -> Option<Option<&Vec<CqlValue>>> {
```
piodul, line 184: We can simplify a bit here, similarly to `is_value_deleted`. I don't think that the set of deleted elements can be empty, so we can consider null to be an empty Vec in order to remove one layer of Option. Additionally, you should change `&Vec<CqlValue>` to `&[CqlValue]` - always prefer slices to vec references.

Also, take a look at my summary comment of this review - I included more suggestions which will affect the final shape of this function.

piodul (changes requested):
Sorry for changing my mind on the interface so late... but after I saw how many unwraps there are in tests I think the return types `get_value`, `is_value_deleted` and `get_deleted_elements` should be simplified even if it means that some information about columns existing or not will be unavailable. I think that the users will usually know what their schema is, so they which column names to expect and which `cdc$deleted_` and `cdc$deleted_elements_` special columns to expect. They will have to unwrap anyway (I don't think they will want to handle it in some other way), so we can do it for them. I imagine the following interface:

```rust
// We need this one `Option` here because it represents if the value is null or not
// and that is a case some users would definitely want to check for
pub fn get_value(&self, name: &str) -> &Option<CqlValue> {
    // Panic if the column does not exist
    // Otherwise return a reference to its value
}

pub fn is_value_deleted(&self, name: &str) -> bool {
    // Panic if the column does not exist
    // Otherwise return if the value was deleted
}

pub fn get_deleted_elements(&self, name: &str) -> &[CqlValue] {
    // Panic if the column does not exist
    // If the value is null, return an empty slice: &[]
    // Otherwise return the slice with deleted elements
}
```

If I'm wrong and somebody does need this information, we can add another method for them which checks if the column exists:

```rust
pub fn column_exists(&self, name: &str) -> bool {
	// ...
}
```

The panicking behavior should be documented in the docstring comments for those methods.

Apart from that and some other small comments I left, I think it LGTM.

piodul:
@kbr- please review, too.

piodul:
@kbr- ping

piodul:
@kbr- ping^2

kbr-:
Please fix the git log according to these instructions:
https://github.com/piodul/scylla-cdc-rust/pull/8#issuecomment-967174755

also the PR cover letter says:
> This is definitely not finished, as you can see, there are many TODOs and it does not compile.

I'm guessing this is no longer true, so please update it.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,474 @@
...
+    PartitionDelete,
+    RowRangeDelInclLeft,
+    RowRangeDelExclLeft,
+    RowRangeDelInclRight,
+    RowRangeDelExclRight,
+    PostImage,
+}
+
+pub struct CDCRowSchema {
+    pub(crate) stream_id: usize,
```
kbr-, line 29: These values are indices of these columns in the CDC log table schema, or something else? A comment should mention this. (I didn't understand what these `usize`s mean until I read the implementation of `new`)

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,474 @@
...
+        let mut end_of_batch = false;
+        let mut operation = OperationType::PreImage;
+        let mut ttl = None;
+
+        let data_count =
+            schema.mapping.len() + schema.deleted_mapping.len() + schema.deleted_el_mapping.len();
+        let mut data: Vec<Option<CqlValue>> = Vec::with_capacity(data_count);
+
+        for (i, column) in row.columns.into_iter().enumerate() {
+            if i == schema.stream_id {
```
kbr-, line 134: Perhaps using guards would be more elegant:
https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html
(just a suggestion, depends on your preference)

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,474 @@
...
+            .get(name)
+            .map(|id| self.data[*id].as_ref().map(|val| val.as_set().unwrap()))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    // Because we are planning to extract a common setup to all tests,
+    // the setup for this module is based on generation fetcher's tests,
+    // which has already been merged to the main branch.
```
kbr-, line 196: The last part of this commit is really unnecessary (about merging stuff into `main`)

kbr- (approved):

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,485 @@
...
+            None => &[],
+        }
+    }
+
+    pub fn column_exists(&self, name: &str) -> bool {
+        self.schema
+            .mapping
+            .get(name)
+            .map(|id| &self.data[*id])
+            .is_some()
```
piodul, line 218: The `map` here is completely unnecessary. It doesn't change the result of the latter `is_some()` in any way, and the `&self.data[*id]` is not used at all.

Instead, you can probably do the following:

```rust
self.schema
    .mapping
    .contains_key(name)
```

The `column_deletable` and `collection_exists` methods have the same problem.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,485 @@
...
+        let val = self
+            .schema
+            .deleted_el_mapping
+            .get(name)
+            .map(|id| self.data[*id].as_ref().map(|val| val.as_set().unwrap()))
+            .unwrap();
+        match val {
+            Some(vec) => vec,
+            None => &[],
+        }
```
piodul, line 210: Nit: this is probably equivalent to `val.unwrap_or(&[])`

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,485 @@
...
+const TTL_NAME: &str = "cdc$ttl";
+const IS_DELETED_PREFIX: &str = "cdc$deleted_";
+const ARE_ELEMENTS_DELETED_PREFIX: &str = "cdc$deleted_elements_";
+
+impl CDCRowSchema {
+    pub fn new(specs: &[ColumnSpec]) -> CDCRowSchema {
+        // [TODO]:  This function is not usable now.
+        //          The driver does not provide the vector of ColumnSpec for every query yet,
+        //          and therefore we can't map the result
+        //          to something other than a hardcoded tuple.
```
piodul, line 63: Please remove this comment. I suspect nobody will bother to remove it later, and it can only cause confusion. I'm not sure it provides any value.

scylla-cdc/src/consumer.rs
```diff
@@ -0,0 +1,485 @@
...
+            .get(name)
+            .map(|id| self.data[*id].as_ref().map(|val| val.as_set().unwrap()))
+            .unwrap();
+        match val {
+            Some(vec) => vec,
+            None => &[],
+        }
+    }
+
+    pub fn column_exists(&self, name: &str) -> bool {
```
piodul, line 213: I think the newly added methods belong to `CDCRowSchema` rather than `CDCRow`, so please move them there.

procrastinationfighter: I thought that these functions are for the consumer so that he can use them, for example, to debug. As far as I know, the consumer isn't going to see any `CDCRowSchema`, because this is an internal struct that is used to create new `CDCRow` objects. If these functions were moved to `CDCRowSchema`, they would be useless in my opinion.

scylla-cdc/src/consumer.rs
procrastinationfighter: I'm not sure about that. It does not compile with message: `expected struct `Vec`, found array of 0 elements`. `val` is of type `Option<&Vec<CqlValue>>` and if I understand the generics correctly, `unwrap_or` should take as an argument the same type as its return type: https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

scylla-cdc/src/consumer.rs
piodul: Personally, I see nothing wrong with exposing the `CDCRowSchema` to the user, (for example through `CDCRow::get_schema()` or something) and I think that those methods make more sense to be there, but maybe it will make the interface harder to use... I'm not sure, so I won't insist.

piodul (approved):

PR #27, opened 2021-12-08
Added support for async operations inside Consumer trait.

### LINKED ISSUE #26: Consumer: Add support for async operations.
Currently, Consumer trait is not async, so things like replication (see #17 ) are impossible to do as per [source](https://stackoverflow.com/questions/66035290/how-do-i-await-a-future-inside-a-non-async-method-which-was-called-from-an-async), since requires using `await` for queries on destination table.

macher259:
Fixes #26

I added support for async operations for Consumer trait by introducing a new dependency `async-trait` that is advised as per [source](https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/) and making Consumer trait async.

### DIFF
--- a/scylla-cdc/Cargo.toml
+++ b/scylla-cdc/Cargo.toml
@@ -13,6 +13,7 @@ chrono = "0.4.19"
 futures = "0.3.17"
 uuid = "0.8.2"
 num_enum = "0.5.4"
+async-trait = "0.1.51"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


piodul (approved):

piodul:
@macher259 One more thing before merge: please correct the title and the commit message - you used the "Customer" word, while it should be "Consumer".

piodul (approved):

Issue #17, opened 2021-12-07
Replicator consumer stub


piodul:
Context: the CDC replicator is an application which replays changes made to a table in one cluster, into another table with the same schema in another cluster.

The idea is to provide a stub implementation of the `Consumer` trait for the replicator. The consumer should have all of the necessary components needed to replicate the data. The actual logic doesn't have to actually replicate all of the data correctly yet, but ideally it should provide a framework so that after merging the stub, the work on supporting different CQL types and operation types can be parallelized.

The consumer should:

- Have an `Arc<Session>` for the destination cluster,
- Keyspace and table name of the destination table,
- Have the information about the schema of the __base__ table (the table which the CDC log is based on) - it should fetch this information automatically from the cluster metadata (this functionality is on master but not yet any release, so we need to bump the dependency to temporarily point at the right commit),
- Support inserts and updates only, can panic if other operation types are encountered,
- Support clustering/static columns with `int` type only, can panic if the table has other clustering/static columns with that type,
- Should make sure that the changes are replicated with the correct timestamp (see the `set_timestamp` method)

We also need a simple test framework for writing replicator tests. This can be a single function which:

- Creates one, (source) table with specified schema with CDC enabled, and another (destination) table with the same schema but without CDC - no need for two clusters, all of those tables can be created in one and the same cluster,
- Takes a list of queries and executes them sequentially,
- Reads all data from the CDC table and feeds it row by row to the replicator consumer,
- Fetches rows from the source and destination table and compares them - they should be equal.
  - As a bonus, the test can also verify that the column timestamps are the same (tip: use the [writetime](https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useWritetime.html) function).

Issue #18, opened 2021-12-07
Replicator: support atomic types in inserts/updates


piodul:
https://docs.scylladb.com/using-scylla/cdc/cdc-basic-operations/#updates
https://docs.scylladb.com/using-scylla/cdc/cdc-basic-operations/#inserts

piodul:
Fixed by #28

PR #28, opened 2022-01-08
Create stub of the replicator.

procrastinationfighter:
Fixes: #17 #18 
This PR adds small changes to the `consumer` module, creates a stub of the data replicator and a framework to test the replicator.
It modifies the `consumer.rs` in order to set stream_id to match the internal type used in reader and generation fetcher and changes `ConsumerFactory`'s `new_consumer` method to be async.
The stub supports only inserts and updates for atomic types and frozen collections/UDTs.


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    async fn update_or_insert(&self, data: CDCRow<'_>, is_insert: bool) {
+        // [TODO]: Does not support deletions yet.
+        let keys_iter = ReplicatorConsumer::get_keys_iter(&self.table_schema).await;
+        let ttl = data.ttl.unwrap_or(0); // If data is inserted without TTL, setting it to 0 deletes existing TTL.
+        let timestamp = 0; // [TODO]: Map uuid to i64
+                           // [FIXME]: CqlValue does not have implemented FromIterator.
+        let mut values = keys_iter
+            .map(|col_name| data.get_value(col_name).unwrap())
+            .collect::<Vec<CqlValue>>();
```
procrastinationfighter, line 118: Small update: due to tiredness, I thought that the problem with `CqlValue` is more complex than it really is. Actually, `Value` trait is implemented for `CqlValue` so our only problem here is that we can't use iterators to map the values. This can be solved by implementing `FromIterator` trait for `CqlValue`. Alternatively, we could just not use the iterators and clone the values by using simple loop, but I think that we need the `Copy` trait, which is AFAIK also not implemented for `CqlValue` (I can't see it [in the documentation](https://docs.rs/scylla/latest/scylla/frame/response/result/enum.CqlValue.html#trait-implementations)). So, my question is: should we add one (or both) of these traits to `CqlValue` and if yes, should it be done in this crate or in the driver?
Bonus question, after we solve this issue: given a `key_values: Vec<CqlValue>`, is there a smart way to create a vector like this one: `[ttl, val] ++ key_values` (where `++` stands for vector concatenation) that neither copies `key_values' nor moves it? The query for value deletion will require different values than an update, so I'm not sure if we can afford to keep two separate vectors, or should merge the vectors as appropriate.

piodul:
I pulled this and tried to compile, but I'm getting the following error:

```
error[E0277]: a value of type `Vec<CqlValue>` cannot be built from an iterator over elements of type `scylla::frame::response::result::CqlValue`
   --> scylla-cdc-replicator/src/replicator_consumer.rs:118:14
    |
118 |             .collect::<Vec<CqlValue>>();
    |              ^^^^^^^ value of type `Vec<CqlValue>` cannot be built from `std::iter::Iterator<Item=scylla::frame::response::result::CqlValue>`
    |
    = help: the trait `FromIterator<scylla::frame::response::result::CqlValue>` is not implemented for `Vec<CqlValue>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `scylla-cdc-replicator` due to previous error
```

This is caused by the version mismatch of the `scylla` crate. `scylla-cdc-replicator` uses version "0.3.0", but `scylla-cdc` uses the master version.

> Actually, Value trait is implemented for CqlValue so our only problem here is that we can't use iterators to map the values. This can be solved by implementing FromIterator trait for CqlValue

While rustc's error messages usually describe the problem very clearly, the case above is one of the rare exceptions. What gives the issue away is the fact that one CqlValue in the error description is fully qualified and the other is not - namely, the unqualified one refers to the one you imported in `replicator_consumer.rs`, and the other one is from another version of the `scylla` library so it is not imported and printed in a fully qualified way.

The FromIterator issue is a red herring and it has probably confused you. `FromIterator` is usually implemented for containers, e.g. Vec or HashMap, and is used to "materialize" a collection from an iterator (this trait is used by the `collect` method) - I doubt it would make sense to implement it for CqlValue.

----

Anyway, if you change "0.3.0" to "master" in `scylla-cdc-replicator` dependencies then other issues pop up - mostly related to the fact that you are trying to construct a `Vec<CqlValue>`, but there is no way to "take out" values from `CDCRow` or clone `CqlValue`s which are available through references. There are three ways to approach this:

1. Add `#[derive(Clone)]` for `CqlValue` (in the rust driver project). We can't implement `Copy` because it can only be implemented for simple values which can be copied via a `memcpy`-like procedure; `CqlValue::Set` for example contains a `Vec` so `Copy` is out of the question. This solution is a bit sad because cloning `CqlValue`s might be sometimes expensive.
1. Add possibility to move values out from a `CDCRow`. This way we could avoid cloning.
1. Operate solely on references to `CqlValue`. Operations such as `Session::execute` or `Session::query` can operate on vectors/slices of _references_ of `CqlValue`s This way we could even avoid moving.

I'll send a PR for 1. because it's trivial.
For 2., I created an issue: https://github.com/piodul/scylla-cdc-rust/issues/29 to track that.

I modified your code locally according to the third strategy and managed to compile it (although with warnings), so there is a good chance that this approach will work - if not, we can fall back to 2.. Here's the diff:

```diff
diff --git a/scylla-cdc-replicator/src/replicator_consumer.rs b/scylla-cdc-replicator/src/replicator_consumer.rs
index 909f5e2..5c994c2 100644
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -114,8 +114,9 @@ impl ReplicatorConsumer {
         let timestamp = 0; // [TODO]: Map uuid to i64
                            // [FIXME]: CqlValue does not have implemented FromIterator.
         let mut values = keys_iter
-            .map(|col_name| data.get_value(col_name).unwrap())
-            .collect::<Vec<CqlValue>>();
+            .clone()
+            .map(|col_name| data.get_value(col_name).as_ref().unwrap())
+            .collect::<Vec<&CqlValue>>();
 
         // [TODO]: Append tll to values and translate values to some format that implements ValueList.
         if is_insert {
@@ -124,7 +125,6 @@ impl ReplicatorConsumer {
         }
 
         let keys_cond = keys_iter
-            .clone()
             .map(|name| {
                 name.to_owned()
                     + " = "
@@ -163,14 +163,14 @@ impl ReplicatorConsumer {
     async fn run_prepared_statement(
         &self,
         mut query: PreparedStatement,
-        values: &[CqlValue],
+        values: &[&CqlValue],
         timestamp: i64,
     ) {
         query.set_timestamp(Some(timestamp));
         self.dest_session.execute(&query, values);
     }
 
-    async fn run_statement(&self, mut query: Query, values: &[CqlValue], timestamp: i64) {
+    async fn run_statement(&self, mut query: Query, values: &[&CqlValue], timestamp: i64) {
         query.set_timestamp(Some(timestamp));
         self.dest_session.query(query, values);
     }
diff --git a/scylla-cdc/Cargo.toml b/scylla-cdc/Cargo.toml
index 7ff86ed..9887215 100644
--- a/scylla-cdc/Cargo.toml
+++ b/scylla-cdc/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0.48"
-scylla = "0.3.0"
+scylla = { git = "https://github.com/scylladb/scylla-rust-driver" }
 tokio = { version = "1.1.0", features = ["rt", "io-util", "net", "time", "macros", "sync"] }
 chrono = "0.4.19"
 futures = "0.3.17"
```

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: > Bonus question, after we solve this issue: given a `key_values: Vec<CqlValue>`, is there a smart way to create a vector like this one: `[ttl, val] ++ key_values` (where `++` stands for vector concatenation) that neither copies `key_values' nor moves it?

Unfortunately, no. Perhaps if `Session::execute` or `Session::query` accepted an iterator then it would be possible (iterators can be chained together)

> The query for value deletion will require different values than an update, so I'm not sure if we can afford to keep two separate vectors, or should merge the vectors as appropriate.

Maybe you can just have two separate code paths which construct the values vector depending on the case?

Tip: if you want to efficiently construct a vector like `[ttl, val] ++ key_values` where `key_values` is an iterator, you can use the `extend` method:

```rust
// You can preallocate with `.reserve()` if you know how many values are in `key_values`
let mut values = vec![ttl, val];
values.extend(key_values);
```

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,237 @@
...
+            dest_table_name,
+            table_schema,
+            other_columns,
+            insert_query,
+        }
+    }
+
+    async fn get_keys_iter(
+        table_schema: &Table,
+    ) -> std::iter::Chain<Iter<'_, String>, Iter<'_, String>> {
```
piodul, line 82: Tip: you can simplify the return type using [`impl Trait`](https://doc.rust-lang.org/reference/types/impl-trait.html):

```rust
) -> impl std::iter::Iterator<Item=String> {
```

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,237 @@
...
+use scylla_cdc::consumer::*;
+use std::slice::Iter;
+use std::sync::Arc;
+
+struct ReplicatorConsumer {
+    dest_session: Arc<Session>,
+    dest_keyspace_name: String,
+    dest_table_name: String,
+    // [TODO]: Not sure, if we want to keep the ownership here,
+    //   or just a ref and keep it in the factory.
```
piodul, line 18: In this case, it's better to either keep the table description by value (`Table`) or via a thread-safe shared pointer (`Arc<Table>`). Our code will spawn multiple consumers and run them on different tokio tasks. It's not really possible to have references pointing "between" tokio tasks.

scylla-cdc-replicator/src/replicator_consumer.rs
procrastinationfighter: I tried to do so, but I think it should iterate over `&String`, not `String`. Nevertheless, the clones don't work with this solution:
```
error[E0599]: no method named `clone` found for opaque type `impl Iterator` in the current scope
  --> scylla-cdc-replicator/src/replicator_consumer.rs:49:31
   |
49 |         let names = keys_iter.clone().join(",");
   |                               ^^^^^ method not found in `impl Iterator`
```

procrastinationfighter:
I applied changes and added support for deleting a value, so I think that this part of the issue is ready to review now. 
I will try to create the framework for testing before Friday, after that I will mark this PR as ready for review.

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: How about `impl std::iter::Iterator<Item=&String> + Clone`, then?

procrastinationfighter: It compiles, I've just pushed this (see below).

piodul:
> I also didn't find a way to map `UUID` to an `i64` timestamp yet.

Some hints:
- As far as I remember, `i64` timestamp must be expressed as a number of microseconds since the unix epoch (it would be good to verify it experimentally),
- Look for conversion functions in the `uuid` crate documentation - the page for the `Uuid` type is a good starting point: https://docs.rs/uuid/0.8.2/uuid/struct.Uuid.html

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,261 @@
...
+                self.run_statement(
+                    Query::new(format!(
+                        "DELETE {} FROM {}.{} WHERE {}",
+                        column_name, self.dest_keyspace_name, self.dest_table_name, self.keys_cond
+                    )),
+                    &values,
+                    timestamp,
+                )
+                .await;
+            }
```
piodul, line 171: This will work, but only for frozen collections/UDTs and atomic types. Please move this logic to a separate function. In this loop, please check the column's type using information in the `table_schema` and call the new function if it matches. This way it will be easier for others to add support for other types in parallel.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,261 @@
...
+    dest_session: Arc<Session>,
+    dest_keyspace_name: String,
+    dest_table_name: String,
+    table_schema: Table,
+}
+
+impl ReplicatorConsumerFactory {
+    /// Creates a new instance of ReplicatorConsumerFactory.
+    /// Fetching schema metadata must be enabled in the session.
+    pub async fn new(
```
piodul, line 224: It doesn't have to be async.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+    name: String,
+    partition_key: Vec<TestColumn>,
+    clustering_key: Vec<TestColumn>,
+    other_columns: Vec<TestColumn>,
+}
+
+// [TODO]: We might want to provide queries for common operations like insert or update with one value.
+type TestOperation = (String, Vec<CqlValue>);
+
+fn get_type_name(col_type: &CqlType) -> &str {
```
piodul, line 19: This function might become large if fully implemented (there are many native types and we would have to handle all of them). Could we switch from `CqlType` in `TestColumn` to just `String`, for simplicity?

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+    session.query("CREATE KEYSPACE IF NOT EXISTS test_dst WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1}", ()).await?;
+
+    let partition_key_name: String = if schema.partition_key.len() == 1 {
+        schema.partition_key[0].0.clone()
+    } else {
+        format!(
+            "({})",
+            schema.partition_key.iter().map(|(name, _)| name).join(",")
+        )
+    };
```
piodul, line 39: If you want to be fancy, you can use slice patterns:

```rust
let partition_key_name: String = match schema.partition_key.as_slice() {
    [pk] => pk.0.clone(),
    _ => format!(
        "({})",
        schema.partition_key.iter().map(|(name, _)| name).join(",")
    ),
};
```

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,248 @@
...
+                timestamp,
+            )
+            .await;
+        }
+    }
+
+    async fn update_or_insert(&self, data: CDCRow<'_>, is_insert: bool) {
+        let keys_iter = ReplicatorConsumer::get_keys_iter(&self.table_schema);
+        let ttl = CqlValue::Int(data.ttl.unwrap_or(0) as i32); // If data is inserted without TTL, setting it to 0 deletes existing TTL.
+        let timestamp = ReplicatorConsumer::get_timestamp(&data);
```
procrastinationfighter, line 126: There was a problem with TTL. When I used `CqlValue:Bigint`, I received the following error:
```
DbError(Invalid, "Exception while binding column \"[ttl]\": marshaling error: Validation failed for type org.apache.cassandra.db.marshal.Int32Type: got 8 bytes")
```
This is strange, because CDC table holds ttl as bigint (which consists of 8 bytes).

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -9,7 +9,7 @@ use scylla::Session;
 use scylla_cdc::consumer::*;
 use std::sync::Arc;
 
-struct ReplicatorConsumer {
+pub(crate) struct ReplicatorConsumer {
```
piodul, line 12: Fixes to this file should do not belong to this commit - this commit is supposed to only add replication_tests. Please move those changes into a separate commit and then squash the new commit with "Create a stub of the replicator."

piodul (changes requested):

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,244 @@
...
+        for column_name in &self.non_key_columns {
+            match self.table_schema.columns.get(column_name).unwrap().type_ {
+                CqlType::Native(_)
+                | CqlType::Tuple(_)
+                | CqlType::Collection { frozen: true, .. }
+                | CqlType::UserDefinedType { frozen: true, .. } => {
+                    self.update_atomic_or_frozen(
+                        column_name,
+                        &data,
+                        values_for_update.clone(), // We can't pass a mutable ref here if we want to run async functions in a loop.
```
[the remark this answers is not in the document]
procrastinationfighter: This is weird. I swear that there was a compillation error saying that I can't use `&mut` because there can be only one mutable reference at a time, and one has been created in the previous iteration of the loop, but maybe it was related to some other bug that was removed in the meantime. Since it works, I'll just remove the comment and use `&mut`.

piodul (approved):

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    fn get_keys_iter(table_schema: &Table) -> impl std::iter::Iterator<Item = &String> + Clone {
+        table_schema
+            .partition_key
+            .iter()
+            .chain(table_schema.clustering_key.iter())
+    }
+
+    fn get_timestamp(data: &CDCRow<'_>) -> i64 {
+        const NANOS_IN_MILLIS: u64 = 1000;
```
kbr-, line 78: this doesn't seem right

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    async fn run_prepared_statement(
+        &self,
+        mut query: PreparedStatement,
+        values: &[&CqlValue],
+        timestamp: i64,
+    ) {
+        query.set_timestamp(Some(timestamp));
+        self.dest_session
+            .execute(&query, values)
```
kbr-, line 174: `execute` and `query` API should probably be extended to take a timestamp param so you don't have to clone the `query` every time

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -0,0 +1,272 @@
...
+        let original_rows = session
+            .query(format!("SELECT * FROM test_src.{}", name), ())
+            .await?
+            .rows
+            .unwrap_or(vec![]);
+        let replicated_rows = session
+            .query(format!("SELECT * FROM test_dst.{}", name), ())
+            .await?
+            .rows
+            .unwrap_or(vec![]);
```
kbr-, line 173: It would be even better for the test to also compare timestamps of at least one column
(you can select the timestamp of a particular column using CQL `writetime(...)` function)

procrastinationfighter: This was meant to be here, but we wanted to merge this PR asap, so we decided to delay this: #31

Issue #29, opened 2022-01-11
Add possibility to move values out of CDCRow


piodul:
Currently, `CDCRow` allows only for borrowing values from it (e.g. `pub fn get_value(&self, name: &str) -> &Option<CqlValue>`), but does not allow to move those values out. Sometimes, users might need to operate on `CqlValue`s directly instead of references, and cloning the value (...which is not yet implemented in the driver...) might be expensive for collections.

We should expose couterparts to our current getters which return an owned value instead of a reference. Here is the proposed interface:

```rust
// Note the lack of ampersand before CqlValue
pub fn take_value(&self, name: &str) -> Option<CqlValue> { /* ... */ }
pub fn take_deleted_elements(&self, name: &str) -> Vec<CqlValue> { /* ... */ }
```

These methods should leave the row in a valid state. Calling them for the second time on the same column can return some dummy value, e.g. `take_value` can return `None` and `take_deleted_elements` can return `vec![]` (this behavior should be documented).

PR #30, opened 2022-01-12
Added possibility to move values out of CDCRow

macher259:
# Fixes: #29 
# Description:
As in issue description:
>Currently, `CDCRow` allows only for borrowing values from it (e.g. `pub fn get_value(&self, name: &str) -> &Option<CqlValue>)`, but does not allow to move those values out. Sometimes, users might need to operate on CqlValues directly instead of references, and cloning the value (...which is not yet implemented in the driver...) might be expensive for collections.

Now users can take values from `CDCRow` leaving `None` in place of the value taken.

### DIFF
--- a/scylla-cdc/src/consumer.rs
+++ b/scylla-cdc/src/consumer.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use num_enum::TryFromPrimitive;
+use scylla::frame::response::result::CqlValue::Set;
 use scylla::frame::response::result::{ColumnSpec, CqlValue, Row};
 use std::collections::HashMap;
 
@@ -215,6 +216,22 @@ impl CDCRow<'_> {
             Some(vec) => vec,
             None => &[],
         }
+    }
+
+    /// Allows to take deleted elements from a collection.
+    /// Returns new empty vector if the value is null or such column doesn't exist.
+    /// The returned value is always owned.
+    /// Leaves None in place of taken data.
+    pub fn take_deleted_elements(&mut self, name: &str) -> Vec<CqlValue> {
+        self.schema
+            .deleted_el_mapping
+            .get(name)
+            .and_then(|id| self.data[*id].take())
+            .and_then(|x| match x {
+                Set(x) => Some(x),
+                _ => None,
+            })
+            .unwrap_or_default()
     }
 
     pub fn column_exists(&self, name: &str) -> bool {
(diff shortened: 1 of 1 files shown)


scylla-cdc/src/consumer.rs
```diff
@@ -207,6 +219,21 @@ impl CDCRow<'_> {
         }
     }
 
+    /// Allows to take deleted elements from a collection.
+    /// Returns new empty vector if the value is null or such column doesn't exist.
+    /// The returned value is always owned.
+    /// Leaves None in place of taken data.
+    pub fn take_deleted_elements(&mut self, name: &str) -> Vec<CqlValue> {
+        let val = self.schema.deleted_el_mapping.get(name).map(|id| {
```
piodul, line 227: If you change `map` to `and_then` here then you can get rid of flatten at the end.

You can also make the whole expression less nested:

```rust
self.schema
    .deleted_el_mapping
    .get(name)
    .and_then(|id| self.data[*id].take())
    .and_then(|x| match x {
        Set(x) => Some(x),
        _ => None,
    })
    .unwrap_or_default()
```

scylla-cdc/src/consumer.rs
```diff
@@ -467,4 +494,57 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_take_deleted_elements() {
+        let session = setup().await.unwrap();
+        session
+            .query(
+                format!(
+                    "UPDATE {} SET vs = vs - {{{}}} WHERE pk = {} AND ck = {}",
+                    TEST_SINGLE_COLLECTION_TABLE, 2, 1, 2
```
piodul, line 524: Don't interpolate query parameters using `format`. Instead, use bind markers (`?`) and provide values in the designated parameter of the `query` method.

It should be possible to convert every parameter here with the exception of the table name.

scylla-cdc/src/consumer.rs
```diff
@@ -467,4 +494,57 @@ mod tests {
...
+                ),
+                (),
+            )
+            .await
+            .unwrap();
+        // We must allow filtering in order to search by cdc$operation.
+        let result = session
+            .query(format!("SELECT * FROM {} WHERE \"cdc$operation\" = {} AND pk = {} AND ck = {} ALLOW FILTERING;",
+                           TEST_SINGLE_COLLECTION_CDC_TABLE, OperationType::RowUpdate as i8, 1, 2), ())
```
piodul, line 533: Same thing - don't use string interpolation for passing parameters.

piodul (changes requested):
Just a couple of minor issues to fix and it will LGTM.

piodul (approved):

PR #34, opened 2022-01-22
Replicator: add support for non-frozen maps in inserts/updates

### LINKED ISSUE #20: Replicator: support non-frozen maps in inserts/updates
https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#maps

macher259:
# Fixes #20 
# Description:
Added support for the simplest collection type - maps in INSERT/UPDATE/DELETE statement. 

Thank to great work done by @procrastinationfighter, the only thing needed was to write tests.


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -144,12 +145,19 @@ impl ReplicatorConsumer {
...
 
         for column_name in &self.non_key_columns {
-            match self.table_schema.columns.get(column_name).unwrap().type_ {
+            match self
+                .table_schema
+                .columns
+                .get(column_name)
+                .unwrap()
+                .type_
+                .clone()
```
piodul, line 154: No need to introduce clone here. You can put `&` before `self.table_schema...` and the type will be matched on by reference.

procrastinationfighter:
I think that the short description of the issue misguided you. Yes, adding, deleting or replacing the whole map works, but the whole point of non-frozen collections is that they can have their content modified (for example, look at the `get_deleted_elements` function in the `CDCRow` structure). 
More info about non-frozen maps: https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#maps

piodul (changes requested):
I'm afraid that it's not that simple. Non-frozen types additionally allow for partial updates, and those operations are represented in the CDC log in a different way. Examples of such operations:

```cql
UPDATE ks.t SET v = v + {1: 'v1', 2: 'v2'} WHERE pk = 0 AND ck = 0;
UPDATE ks.t SET v = v - {1, 2, 3} WHERE pk = 0 AND ck = 0;
```

Please refer to the documentation about how "advanced types" are represented in CDC ([link](https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#maps)) and make sure that partial updates work, too.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -317,4 +317,54 @@ mod tests {
...
+                    Map(vec![(Int(10), Int(-10))]),
+                    Map(vec![(Int(11), Boolean(false))]),
+                ],
+            ),
+            (
+                "UPDATE MAPS_UPDATE SET v1 = v1 - ? WHERE pk = ? AND ck = ?",
+                vec![Set(vec![Int(10)]), Int(10), Int(20)],
+            ),
+            (
+                "UPDATE MAPS_UPDATE SET v2 = v2 + ? WHERE pk = ? AND ck = ?",
```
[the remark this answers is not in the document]
macher259: Yeah, it was working, but it was missing in tests. I have added a single test for overwriting and another one for deleting maps.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -87,6 +87,33 @@ impl ReplicatorConsumer {
...
+        values_for_delete: &[&CqlValue],
+        timestamp: i64,
+    ) {
+        let deleted_elements: &'a [CqlValue] = data.get_deleted_elements(column_name);
+
+        if (data.get_value(column_name).is_none() && deleted_elements.is_empty())
+            || data.is_value_deleted(column_name)
+        {
+            // INSERT/DELETE/OVERWRITE
+            self.update_atomic_or_frozen(
```
kbr-, line 162: That's very confusing, since both the name and comment on `update_atomic_or_frozen` suggest that it can only be used for atomic/frozen types, and here you're using it for a non-frozen collection.

The function's contract should be improved (generalized).

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -87,6 +87,33 @@ impl ReplicatorConsumer {
...
+        column_name: &str,
+        data: &'a CDCRow<'_>,
+        values_for_update: &mut Vec<&'a CqlValue>,
+        values_for_delete: &[&CqlValue],
+        timestamp: i64,
+    ) {
+        let deleted_elements: &'a [CqlValue] = data.get_deleted_elements(column_name);
+
+        if (data.get_value(column_name).is_none() && deleted_elements.is_empty())
+            || data.is_value_deleted(column_name)
```
kbr-, line 159: Please explain the reasoning behind the condition in a comment.

kbr-: It seems to me that the second part of the alternative is sufficient. INSERT and DELETE both overwrite the collection, inserting a collection-wide tombstone, so `is_value_deleted` is true.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -317,4 +317,62 @@ mod tests {
...
+                ],
+            ),
+            (
+                "UPDATE MAPS_UPDATE SET v2 = ? WHERE pk = ? AND ck = ?",
+                vec![Map(vec![(Int(2), Boolean(true))]), Int(10), Int(20)],
+            ),
+            (
+                "DELETE v1 FROM MAPS_UPDATE WHERE pk = ? AND ck = ?",
+                vec![Int(1), Int(2)],
+            ),
```
kbr-, line 371: Putting overwrite statements last make the test less accurate. If replicating overwriting statements works but adding/deleting elements doesn't (`v = v + ?` etc), the test will not catch it (you overwrote both columns which you previously added/deleted elements to/from). BTW. The `INSERT`s at the beginning are not necessary in this test.

You probably need multiple test cases to check every type of statement. Or maybe there is some order of statements such that every statement matters - in this test case right now, only the last two statements do.

Alternatively, we could modify `test_replication` so it compares the tables after every operation, not only at the end.

Finally, it would be great to use property based testing so there's no need to write these tedious test cases by hand.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -110,6 +167,19 @@ impl ReplicatorConsumer {
...
+                column_name,
+                values_for_update,
+                deleted_elements,
+                timestamp,
+            )
+            .await;
+        } else {
+            // V = V + {}
+            self.extend_map(column_name, data, values_for_update, timestamp)
+                .await;
```
kbr-, line 182: These two cases are not mutually exclusive:
```cql
cqlsh> create table ks.t (pk int, ck int, v set<int>, primary key (pk, ck));
cqlsh> update ks.t set v = {1} where pk = 0 and ck = 0;
cqlsh> update ks.t set v = v - {1}, v = v + {2} where pk = 0 and ck = 0;
cqlsh> select * from ks.t;

 pk | ck | v
----+----+-----
  0 |  0 | {2}

```
(using sets in the example, but for maps it's the same)

kbr- (changes requested):

scylla-cdc-replicator/src/replication_tests.rs
kbr-: Now I noticed you're using different keys for the overwrites than for the additions/deletions, so I guess it's OK.

scylla-cdc-replicator/src/replicator_consumer.rs
kbr-: Also, `update_atomic_or_frozen` will replicate collection-wide tombstones using `DELETE` statements, which will use a wrong timestamp. You need to either adjust the timestamp, or change `update_atomic_or_frozen` to replicate deletes using `UPDATE` (`SET v = null`).

Read this:
https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#collection-wide-tombstones-and-timestamps

scylla-cdc-replicator/src/replicator_consumer.rs
macher259: I didn't think of this. Added this in a test case.

scylla-cdc-replicator/src/replicator_consumer.rs
macher259: Yeah, you are right.

scylla-cdc-replicator/src/replicator_consumer.rs
macher259: I have implemented second option and I think it should be correct now.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -87,7 +88,45 @@ impl ReplicatorConsumer {
...
+    // Chained operations are supported.
+    async fn update_map_elements<'a>(
+        &self,
+        column_name: &str,
+        data: &'a CDCRow<'_>,
+        values_for_update: &mut Vec<&'a CqlValue>,
+        timestamp: i64,
+    ) {
+        if let Some(value) = data.get_value(column_name) {
+            // Order of values: ttl, inserted value, pk condition values.
```
kbr-, line 102: s/inserted value/added elements/

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -317,4 +317,87 @@ mod tests {
...
+                    Map(vec![
+                        (Int(-21374134), Boolean(true)),
+                        (Int(-43142137), Boolean(false)),
+                    ]),
+                    Int(1),
+                    Int(2),
+                ],
+            ),
+            (
+                "UPDATE MAP_ELEMENTS_UPDATE SET v1 = v1 - ? WHERE pk = ? AND ck = ?",
```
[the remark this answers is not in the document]
macher259: Yeah, it didn't work. Thanks for pointing this out. I have fixed the issue and rewrote the test a little.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -87,7 +88,72 @@ impl ReplicatorConsumer {
...
+
+        self.run_statement(
+            Query::new(format!(
+                "UPDATE {}.{} USING TTL ? SET {} = {} + ?, {} = {} - ? WHERE {}",
+                self.dest_keyspace_name,
+                self.dest_table_name,
+                column_name,
+                column_name,
+                column_name,
+                column_name,
```
piodul, line 120: You can avoid duplication like this:

```rust
Query::new(format!(
    "UPDATE {ks}.{tbl} USING TTL ? SET {cname} = {cname} + ?, {cname} = {cname} - ? WHERE {cond}",
    ks = self.dest_keyspace_name,
    tbl = self.dest_table_name,
    cond = self.keys_cond,
    cname = column_name,
)),
```

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -91,7 +92,44 @@ impl ReplicatorConsumer {
...
+        data: &'a CDCRow<'_>,
+        values_for_update: &mut Vec<&'a CqlValue>,
+        timestamp: i64,
+    ) -> anyhow::Result<()> {
+        let empty_map = Map(vec![]); // We have to declare it, because we can't take a reference to tmp value in unwrap.
+        let value = data.get_value(column_name).as_ref().unwrap_or(&empty_map);
+        // Order of values: ttl, added elements, pk condition values.
+
+        let deleted_set = Set(Vec::from(data.get_deleted_elements(column_name)));
+        let mut values_for_update: Vec<&CqlValue> = values_for_update.clone();
```
piodul, line 110: That's a shame we have to clone here... However optimizing this will require a refactor too large for this PR, so I guess it's OK for now. It would be nice to do it eventually, but not until all planned replicator consumer features are implemented.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -91,7 +92,44 @@ impl ReplicatorConsumer {
...
+        column_name: &str,
+        data: &'a CDCRow<'_>,
+        values_for_update: &mut Vec<&'a CqlValue>,
+        timestamp: i64,
+    ) -> anyhow::Result<()> {
+        let empty_map = Map(vec![]); // We have to declare it, because we can't take a reference to tmp value in unwrap.
+        let value = data.get_value(column_name).as_ref().unwrap_or(&empty_map);
+        // Order of values: ttl, added elements, pk condition values.
+
+        let deleted_set = Set(Vec::from(data.get_deleted_elements(column_name)));
```
piodul, line 109: Another opportunity for later: here we are creating a vector from a slice, which is also a form of copying. We could take a mutable reference to the CDCRow and take deleted elements out of it.

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: (to be clear: the main priority is to get the replicator to work correctly)

piodul (approved):
LGTM, but I'll wait for @kbr- 's acceptance before merging.

kbr- (approved):

Issue #31, opened 2022-01-21
Replicator: Compare timestamps in the tests


procrastinationfighter:
Quoting from #17 

> As a bonus, the test can also verify that the column timestamps are the same (tip: use the [writetime](https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useWritetime.html) function).

A failed test should print adequate message (see how it's done for other failures).

PR #36, opened 2022-01-22
Replicator: add compare timestamps in the tests

macher259:
# Fixes: #31 
# Description:

This PR adds comparing timestamps of all non-partition & non-clustering columns for all rows.
Edit: Since cql `writetime` function doesn't work for non-frozen collections, we skip them and compare only native types and frozen collections.

### DIFF
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -147,7 +147,9 @@ mod tests {
                 "Number of rows not matching. Original table: {} rows, replicated table: {} rows.",
                 o, r
             ),
-            FailureReason::RowNotMatching(n) => eprintln!("Row {} is not equal in both tables.", n),
+            FailureReason::RowNotMatching(n) => {
+                eprintln!("Row {} is not equal in both tables.", n + 1)
+            }
             FailureReason::TimestampsNotMatching(row, column) => eprintln!(
                 "Timestamps were not equal for column {} in row {}.",
                 column,
@@ -189,7 +191,7 @@ mod tests {
                         name,
                         &original_rows,
                         &replicated_rows,
-                        FailureReason::RowNotMatching(i + 1),
+                        FailureReason::RowNotMatching(i),
                     );
                 }
             }
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -2,8 +2,9 @@
 mod tests {
     use crate::replicator_consumer::ReplicatorConsumer;
     use anyhow::anyhow;
+    use futures_util::FutureExt;
     use itertools::Itertools;
-    use scylla::frame::response::result::CqlValue::{Boolean, Int, List, Map, Set};
+    use scylla::frame::response::result::CqlValue::{Boolean, Int, List, Map, Set, Tuple};
     use scylla::frame::response::result::{CqlValue, Row};
     use scylla::{Session, SessionBuilder};
     use scylla_cdc::consumer::{CDCRow, CDCRowSchema, Consumer};
@@ -29,6 +30,7 @@ mod tests {
     enum FailureReason {
         WrongRowsCount(usize, usize),
         RowNotMatching(usize),
+        TimestampsNotMatching(usize, String),
     }
 
     async fn setup_tables(session: &Session, schema: &TestTableSchema<'_>) -> anyhow::Result<()> {
@@ -146,6 +148,11 @@ mod tests {
                 o, r
             ),
             FailureReason::RowNotMatching(n) => eprintln!("Row {} is not equal in both tables.", n),
+            FailureReason::TimestampsNotMatching(row, column) => eprintln!(
+                "Timestamps were not equal for column {} in row {}.",
+                column,
+                row + 1
+            ),
         }
 
         eprintln!("ORIGINAL TABLE:");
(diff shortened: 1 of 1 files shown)


scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -145,6 +147,10 @@ mod tests {
                 o, r
             ),
             FailureReason::RowNotMatching(n) => eprintln!("Row {} is not equal in both tables.", n),
+            FailureReason::TimestampsNotMatching(row, column) => eprintln!(
+                "Timestamps were not equal for column {} in row {}.",
```
kbr-, line 151: The code below seems to assign numbers to rows starting from 1, IIUC here you start from 0. (should print `row + 1`)

macher259:
Setting this PR to draft since cql `writetime` function used for reading timestamps doesn't support non-frozen collections. Currently replicator doesn't support non-frozen collections, but sooner or later it will.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -198,6 +203,91 @@ mod tests {
...
+            "time",
+            "timestamp",
+            "timeuuid",
+            "tinyint",
+            "uuid",
+            "varint",
+        ];
+
+        for (col_name, col_type) in &schema.other_columns {
+            if NATIVE_TYPES.contains(col_type) || col_type.contains("frozen<") {
```
kbr-, line 270: `map<int, frozen<set<int>>>`
`tuple<int, int>`

scylla-cdc-replicator/src/replication_tests.rs
kbr-: Do you agree with this comment? If yes: it's still printing `row`.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -450,4 +549,49 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn test_checking_timestamps() {
+        let schema = TestTableSchema {
+            name: "COMPARE_TIME".to_string(),
+            partition_key: vec![("pk", "int")],
+            clustering_key: vec![("ck", "int")],
+            other_columns: vec![("v1", "int"), ("v2", "boolean")],
```
kbr-, line 560: I was really hoping that the test would also check some more interesting types like tuples, nested collections (after all there was a bug with nested collections in the initial version of this PR), and also non-frozen collections (to verify that `compare_timestamps` doesn't fail if there is a non-frozen collection column but correctly skips it).

kbr- (changes requested):
Also needs rebasing

scylla-cdc-replicator/src/replication_tests.rs
macher259: I agree with this comment, but I don't see a problem. It shows a correct row number in the test `test_checking_timestamps`, where failing for not matching timestamps is tested. 
`FailureReason::TimestampsNotMatching` in function `compare_timestamps` is constructed as `FailureReason::TimestampsNotMatching(i + 1, col_name.to_string())`, so this `row` here is in fact `row+1`.

scylla-cdc-replicator/src/replication_tests.rs
macher259: My bad. I have added a quick test with those.

scylla-cdc-replicator/src/replication_tests.rs
piodul: I think the `TimestampsNotMatching` should keep a zero-based row index. In Rust, indexing is zero-based, so unless it is documented, the row index being one-based might be confusing. It's better to stick to the convention.

Another problem is that, given that most indexes are zero-based, usually you will have to remember to add `1` every time you construct `TimestampsNotMatching` and it's easy to forget that. If you switch to zero-based index there will be a smaller chance of error and adding `1` will only be a concern of the logic printing the error report.

macher259: > the row index being one-based might be confusing.

I don't think that's the case since rows of the faulty column are enumerated after the main error message, so whoever will look into these errors should see that indexing is one-based. Example of an error message:
```
Replication test for table COMPARE_TIME failed.
Failure reason: Timestamps were not equal for column v2 in row 2.
ORIGINAL TABLE:
Row 1: [Some(BigInt(1645814895337319))]
Row 2: [Some(BigInt(1645814895378103))]
REPLICATED TABLE:
Row 1: [Some(BigInt(1645814895337319))]
Row 2: [Some(BigInt(1645814895338050))]
```

piodul: I meant that using a one-based index _in the code_ is confusing; you should store a zero-based index inside the `TimestampsNotMatching` variant. The conversion to one-based index should be made just before printing.

macher259: Okay, I have changed this along your suggestion.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -562,4 +664,70 @@ mod tests {
...
+                "UPDATE COMPARE_TIME SET v2 = ? WHERE pk = ? AND ck = ?",
+                (false, 1, 2),
+            )
+            .await
+            .unwrap();
+        compare_timestamps(&session, &schema).await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_skipping_non_readable_timestamps() {
```
kbr-, line 714: The test also contains types which do have readable timestamps, e.g. `frozen<set<frozen<set<frozen<set<int>>>>>>` and `tuple<map<int, int>>`. Note that if you nest a collection type in a `tuple`, the entire thing is frozen because `tuple`s cannot be non-frozen.

I like that the test contains all kinds of types, but the test name is confusing - some columns are readable.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -562,4 +664,70 @@ mod tests {
...
+                ("v2", "map<int, int>"),
+                ("v3", "set<frozen<map<int, int>>>"),
+                ("v4", "tuple<map<int, int>>"),
+                ("v5", "frozen<set<frozen<set<frozen<set<int>>>>>>"),
+                ("v6", "tuple<tuple<map<int, int>, set<int>>, int>"),
+                ("v7", "set<frozen<tuple<set<int>, map<int, int>>>>"),
+            ],
+        };
+
+        // Reading timestamps is dependent only on table's schema. This is why we are testing empty columns.
```
kbr-, line 730: I don't understand this comment. In order for `compare_timestamps` to do any work, there must be a row present with live cells. Right now there are no operations so there is nothing to do.

I think you should add some updates so that the table has live cells in the end in each column. Then `compare_timestamps` will compare the timestamps of cells in frozen columns and skip the timestamps in non-frozen columns.

macher259: My intention for this particular test was to test whether our usage of the CQL `WRITETIME` function would result in an error for more complicated types. You are right that no comparison takes place in this test, but the important bit for this test—getting timestamps, does.

In short, comparing timestamps works like this:
```
for every non-pk/ck column:
    if can read timestamp(column):
       original, replicated := get timestamps(column)
       compare timestamps(original, replicated)
 ```
To get timestamps, I use `SELECT WRITETIME(column_name) FROM ...`. Using `WRITETIME` function on non-frozen collections doesn't work and returns an error: `Error message: Cannot use selection function writeTime on non-frozen collections` **no matter if the column in question has any rows**, so even on an empty column it would result in this error. This is the reason I have written this comment.

File: scylla-cdc-replicator/src/replication_tests.rs (hunk 4 above)
Comment (macher259) [reply scylla-cdc-replicator/src/replication_tests.rs]: I have changed this to `test_whether_writetime_function_fails_for_complicated_types`. I hope it's better now.

File: scylla-cdc-replicator/src/replication_tests.rs (hunk 5 above)
Comment (kbr-) [reply scylla-cdc-replicator/src/replication_tests.rs:730]: Ok, makes sense.
On the other hand, we still want to test if the function behaves correctly when/if there are actual cells present.

And my point about frozen columns still stands - we still want to check the correctness of timestamp comparison for `tuple` and `frozen<...>` types, and we need live cells for that.

If you want to have a test with no data that's fine, but then please make a separate test that does have live data **and** complex types (both frozen and non-frozen).

Comment (macher259) [reply scylla-cdc-replicator/src/replication_tests.rs]: I have added insertion for every column and a single element-wise update and tuple delete to this test and renamed it to `test_compare_time_for_complicated_types`. I think that should be enough.

**Review by kbr- — APPROVED** [review]
LGTM.

File: scylla-cdc-replicator/src/replication_tests.rs [hunk 6]
```diff
@@ -600,7 +702,7 @@ mod tests {
 
         let operations = vec![
             (
-                "INSERT INTO PARTITION_DELETE_MULT_PK (pk1, pk2, ck, v1, v2) VALUES (?, ?, ?, ?, ?)",
+                                "INSERT INTO PARTITION_DELETE_MULT_PK (pk1, pk2, ck, v1, v2) VALUES (?, ?, ?, ?, ?)",
```
piodul, line 705: Please remove this indent.

piodul (changes requested):
Some small issues left but otherwise LGTM.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -617,4 +719,142 @@ mod tests {
             .await
             .unwrap();
     }
+
+    #[tokio::test]
+    #[should_panic]
```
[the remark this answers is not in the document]
macher259: The problem is that function for comparing timestamps doesn't return an error, but prints error message and panics. I have changed `#[should_panic]` to asserting that only the tested function panics as per [this SO post](https://stackoverflow.com/questions/26469715/how-do-i-write-a-rust-unit-test-that-ensures-that-a-panic-has-occurred).

piodul (approved):

PR #43, opened 2022-02-16
Replicator: add support for inserting and updating non-frozen list columns.

### LINKED ISSUE #19: Replicator: support non-frozen lists in inserts/updates
https://docs.scylladb.com/using-scylla/cdc/cdc-advanced-types/#lists

procrastinationfighter:
Fixes #19 
This PR adds support for updating and inserting operations with non-frozen lists.
It also includes a small refactor for query responsible for deleting a row.

### DIFF
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -22,6 +22,7 @@ pub(crate) struct ReplicatorConsumer {
     // Prepared queries.
     insert_query: PreparedStatement,
     partition_delete_query: PreparedStatement,
+    delete_query: PreparedStatement,
 
     // Strings for queries created dynamically:
     keys_cond: String,
@@ -73,6 +74,14 @@ impl ReplicatorConsumer {
             .await
             .expect("Preparing partition delete query failed.");
 
+        let delete_query = dest_session
+            .prepare(format!(
+                "DELETE FROM {}.{} WHERE {}",
+                dest_keyspace_name, dest_table_name, keys_cond
+            ))
+            .await
+            .expect("Preparing delete query failed.");
+
         ReplicatorConsumer {
             dest_session,
             dest_keyspace_name,
@@ -81,6 +90,7 @@ impl ReplicatorConsumer {
             non_key_columns,
             insert_query,
             partition_delete_query,
+            delete_query,
             keys_cond,
         }
     }
@@ -211,15 +221,8 @@ impl ReplicatorConsumer {
     async fn delete_row(&self, data: CDCRow<'_>) -> anyhow::Result<()> {
         let (_, timestamp, values) = self.get_common_cdc_row_data(&data);
 
-        self.run_statement(
-            Query::new(format!(
-                "DELETE FROM {}.{} WHERE {}",
-                self.dest_keyspace_name, self.dest_table_name, self.keys_cond
-            )),
-            &values,
-            timestamp,
-        )
-        .await?;
+        self.run_prepared_statement(self.delete_query.clone(), &values, timestamp)
+            .await?;
 
         Ok(())
     }
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -5,6 +5,7 @@ use async_trait::async_trait;
 use itertools::Itertools;
 use scylla::frame::response::result::CqlValue;
 use scylla::frame::response::result::CqlValue::{Map, Set};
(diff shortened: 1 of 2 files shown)


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -157,6 +167,70 @@ impl ReplicatorConsumer {
...
+                self.run_statement(add_query.clone(), values_for_update, timestamp)
+                    .await?;
+            }
+        }
+        let remove_query = Query::new(
+            format!("UPDATE {ks}.{tbl} USING TTL ? SET {list}[SCYLLA_TIMEUUID_LIST_INDEX(?)] = null WHERE {cond}",
+                             ks = self.dest_keyspace_name,
+                             tbl = self.dest_table_name,
+                             list = column_name,
+                             cond = self.keys_cond));
```
procrastinationfighter, line 222: I tried to resolve these two actions by using one query (the `add_query`), but I couldn't pass `null` as `CqlValue` - when I set the value of `SET l[(...)] = ?` to `CqlValue::Empty`, I get the following error:
```
Database returned an error: Invalid protocol message received from the driver, Error message: truncated frame: expected 589829 bytes, length is 6
```
This is why there are two separate queries and two separate lists of arguments.

piodul: `CqlValue::Empty` is not the same as `null`. To represent a nullable value, you can use `Option<CqlValue>`.

procrastinationfighter: Wouldn't I have to replace the type of the vector to `Vec<Option<CqlValue>>` then? Seems like trouble, not sure which version is better.
Btw. if `CqlValue::Empty` is not used for representing `null`, what is its purpose? I couldn't find anything in the documentation.

piodul: `CqlValue::Empty` is a weird case in the Cassandra's/Scylla's type system. Most types, even if they are not conventionally a type that can be empty, can be set to an empty value. For example: `int`. An empty `int` is different from `null` and from `0`.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -167,6 +167,70 @@ impl ReplicatorConsumer {
...
+            )
+            .await?;
+        }
+
+        let add_query = Query::new(
+            format!("UPDATE {ks}.{tbl} USING TTL ? SET {list}[SCYLLA_TIMEUUID_LIST_INDEX(?)] = ? WHERE {cond}",
+                             ks = self.dest_keyspace_name,
+                             tbl = self.dest_table_name,
+                             list = column_name,
+                             cond = self.keys_cond));
```
piodul, line 206: The indentation is a bit strange here. Why not indent it as you did e.g. for the first invocation of `self.run_statement` in this function (`update_list`)?

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: I don't think that changing `values_for_update` to `Vec<Option<CqlValue>>` would be troublesome here, would it? You would also have to change `values_for_list_update`, but it is only used for lists so this won't affect other functions. If you do it, you will also be able to drop the `values_for_remove` argument.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -167,6 +168,68 @@ impl ReplicatorConsumer {
         Ok(())
     }
 
+    // Recreates INSERT/DELETE/UPDATE statement on non-frozen list.
+    async fn update_list<'a>(
+        &self,
+        column_name: &str,
+        data: &'a CDCRow<'_>,
+        values_for_update: &mut Vec<Option<&'a CqlValue>>,
+        values_for_delete: &[&CqlValue],
```
kbr-, line 177: Hmmm, why is this called `values_for_delete`? From the code it seems that the values passed here are primary key values. And that's how they are used too - for the placeholders in the key condition.

Is the variable name as such because it is used in a `DELETE` statement?
It's quite confusing, the parameter names should be chosen so that the caller of the function can understand them - they are part of the interface. They shouldn't be named based on the internal usage by the implementation of that function, IMO. How does the caller know to put primary key into `values_for_delete`?

kbr- (approved):
LGTM.

For the future:
1. Don't rebase and apply changes in a single force-push. Either rebase first and force push, then apply changes and force push, or the other way around. Then the force-push diff is not polluted with rebase stuff.
2. When applying changes, post a comment with a summary of the changes.

Issue #12, opened 2021-11-12
(Very) minimal library for reading CDC


piodul:
Depends on #4 and #5.

After we have a CDC log printer and an interface for consuming changes, those two should be combined - the printer should be extended so that, instead of just printing changes, it accepts user-defined callbacks and calls them on each received change.

In the meantime, the printer application should be reworked so that it uses the new interface to print rows.

PR #46, opened 2022-03-02
Create base for e2e tests.

### LINKED ISSUE #44: Create black-box end-to-end library test
Create a test that works with a Scylla cluster and does the following:

1. Generates a list of operations
2. Registers a consumer with a scylla-cdc-rust library that logs all changes it consumes maintaining the order in which they were consumed.
3. Send all the operations from (1) to Scylla server
4. Check that the operations consumed in (2) are the same as list from (1). That means no operation was missed and all operations were consumed in the same order as they were issued to the server.

procrastinationfighter:
Fixes #44 
This PR adds a base for creating e2e tests that check if operations in CDC are consumed in a proper order. 
As for now, some things are simplified:
- Only Insert and Update operations are supported.
- The table has the following schema: `(pk <PK_TYPE>, ck int, v int, primary key (pk, ck))` where `<PK_TYPE>` is the type of the partition key. Support for complex partition keys might be added later.

This is because these tests should only check the order of the operations and the only thing that may have an impact on it is the partition key. 

In addition, bugs regarding the `StreamReader` have been found and fixed thanks to these tests.

### DIFF
--- a/scylla-cdc/src/e2e_tests.rs
+++ b/scylla-cdc/src/e2e_tests.rs
@@ -1,0 +1,461 @@
+#[cfg(test)]
+mod tests {
+    use crate::cdc_types::ToTimestamp;
+    use crate::consumer::*;
+    use crate::log_reader::CDCLogReader;
+    use anyhow::{anyhow, Result};
+    use async_trait::async_trait;
+    use itertools::Itertools;
+    use scylla::frame::response::result::CqlValue;
+    use scylla::frame::value::{Value, ValueTooBig};
+    use scylla::prepared_statement::PreparedStatement;
+    use scylla::{Session, SessionBuilder};
+    use std::collections::{HashMap, VecDeque};
+    use std::hash::Hash;
+    use std::sync::Arc;
+    use std::time;
+    use tokio::sync::Mutex;
+
+    const SECOND_IN_MILLIS: u64 = 1_000;
+    const SLEEP_INTERVAL: u64 = SECOND_IN_MILLIS / 10;
+    const WINDOW_SIZE: u64 = SECOND_IN_MILLIS / 10 * 3;
+    const SAFETY_INTERVAL: u64 = SECOND_IN_MILLIS / 10;
+
+    type OperationsMap = Arc<Mutex<HashMap<Vec<PrimaryKeyValue>, VecDeque<Operation>>>>;
+
+    // The driver's CqlValue cannot be used as HashMap key,
+    // because it doesn't have the Eq trait.
+    #[derive(Debug, Eq, PartialEq, Hash)]
+    enum PrimaryKeyValue {
+        // Name consistency with CqlValue from the driver is recommended.
+        Int(i32),
+        Text(String),
+        List(Vec<PrimaryKeyValue>),
+    }
+
+    impl Value for PrimaryKeyValue {
+        fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
+            self.to_cql().serialize(buf)?;
+
+            Ok(())
+        }
+    }
+
+    impl PrimaryKeyValue {
+        pub fn from_cql(cql_val: &CqlValue) -> Option<PrimaryKeyValue> {
+            match cql_val {
+                CqlValue::Int(x) => Some(PrimaryKeyValue::Int(*x)),
+                CqlValue::Text(s) => Some(PrimaryKeyValue::Text(s.clone())),
+                CqlValue::List(v) => v
+                    .iter()
+                    .map(PrimaryKeyValue::from_cql)
+                    .collect::<Option<Vec<PrimaryKeyValue>>>()
+                    .map(PrimaryKeyValue::List),
+                _ => None,
+            }
+        }
+
+        pub fn to_cql(&self) -> CqlValue {
+            match self {
+                PrimaryKeyValue::Int(x) => CqlValue::Int(*x),
+                PrimaryKeyValue::Text(s) => CqlValue::Text(s.clone()),
+                PrimaryKeyValue::List(v) => {
+                    CqlValue::List(v.iter().map(PrimaryKeyValue::to_cql).collect())
+                }
+            }
+        }
+    }
+
+    #[derive(Debug, Eq, PartialEq)]
+    struct Operation {
+        operation_type: OperationType,
+        clustering_key: Option<i32>,
+        value: Option<i32>,
+    }
+
+    impl Operation {
+        fn new(
+            operation_type: OperationType,
+            clustering_key: Option<i32>,
+            value: Option<i32>,
+        ) -> Operation {
+            Operation {
+                operation_type,
+                clustering_key,
+                value,
+            }
+        }
+    }
+
+    struct TestConsumer {
+        read_operations: OperationsMap,
+    }
+
+    #[async_trait]
+    impl Consumer for TestConsumer {
+        async fn consume_cdc(&mut self, mut data: CDCRow<'_>) -> Result<()> {
+            let pk_val = {
+                // Primary key columns have names pk1, pk2...
+                let mut values = Vec::new();
+                let mut i = 1;
+                while data.column_exists(&format!("pk{}", i)) {
+                    let val = data.get_value(&format!("pk{}", i)).as_ref().unwrap();
+                    values.push(PrimaryKeyValue::from_cql(val).unwrap());
+                    i += 1;
+                }
+
+                values
+            };
+            let op_type = data.operation.clone();
+            let ck = match data.take_value("ck") {
+                Some(CqlValue::Int(x)) => Some(x),
+                None => None,
+                Some(cql) => return Err(anyhow!(format!("Unexpected ck type: {:?}", cql))),
+            };
+            let val = data.take_value("v").map(|cql| cql.as_int().unwrap());
+
+            self.read_operations
+                .lock()
+                .await
+                .entry(pk_val)
+                .or_insert_with(VecDeque::new)
+                .push_back(Operation::new(op_type, ck, val));
+
+            Ok(())
+        }
+    }
+
+    struct TestConsumerFactory {
+        read_operations: OperationsMap,
+    }
+
+    impl TestConsumerFactory {
+        fn new(operations: OperationsMap) -> TestConsumerFactory {
+            TestConsumerFactory {
+                read_operations: operations,
+            }
+        }
+    }
+
+    #[async_trait]
+    impl ConsumerFactory for TestConsumerFactory {
+        async fn new_consumer(&self) -> Box<dyn Consumer> {
+            Box::new(TestConsumer {
+                read_operations: Arc::clone(&self.read_operations),
+            })
+        }
+    }
+
+    fn get_uri() -> String {
+        std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string())
+    }
+
+    async fn get_session() -> Result<Session> {
+        Ok(SessionBuilder::new().known_node(get_uri()).build().await?)
+    }
+
+    // Creates test table and keyspace. Query should not contain keyspace name, only the table name.
+    async fn create_table_and_keyspace(
+        session: &Session,
+        table_name: &str,
+        pk_type_names: Vec<&str>,
+    ) -> Result<(PreparedStatement, PreparedStatement)> {
+        session.query("CREATE KEYSPACE IF NOT EXISTS e2e_test WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }", ()).await.unwrap();
+        session.await_schema_agreement().await.unwrap();
+        session.use_keyspace("e2e_test", false).await.unwrap();
+
+        session
+            .query(format!("DROP TABLE IF EXISTS {}", table_name), ())
+            .await
+            .unwrap();
+        session.await_schema_agreement().await.unwrap();
+
+        let (create_query, insert_query, update_query) = {
+            let pk_definitions = pk_type_names
+                .iter()
+                .enumerate()
+                .map(|(i, type_name)| format!("pk{} {}", i + 1, type_name))
+                .join(", ");
+            let primary_key_tuple = pk_type_names
+                .iter()
+                .enumerate()
+                .map(|(i, _)| format!("pk{}", i + 1))
+                .join(", ");
+            let binds = pk_type_names.iter().map(|_| "?").join(", ");
+            let pk_conditions = pk_type_names
+                .iter()
+                .enumerate()
+                .map(|(i, _)| format!("pk{} = ?", i + 1))
+                .join(" AND ");
+
+            (
+                format!(
+                    "CREATE TABLE {} ({}, ck int, v int, primary key (({}), ck)) WITH cdc = {{'enabled' : true}}",
+                    table_name, pk_definitions, primary_key_tuple
+                ),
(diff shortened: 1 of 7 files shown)


scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,233 @@ pub mod stream_generations;
...
+            let mut consumer = factory.new_consumer().await;
+
+            let mut rows_stream = self
+                .session
+                .query_iter(
+                    format!("SELECT * FROM {}_scylla_cdc_log", self.table_name),
+                    (),
+                )
+                .await?;
+            let first = rows_stream.next().await.unwrap()?;
```
piodul, line 202: Why do you need to get the first row manually?

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,233 @@ pub mod stream_generations;
...
+    use scylla::{Session, SessionBuilder};
+    use std::collections::{HashMap, VecDeque};
+    use std::hash::Hash;
+    use std::marker::{Send, Sync};
+    use std::sync::Arc;
+    use tokio::sync::Mutex;
+
+    async fn fail_test(mess: &str) {
+        // [TODO]: Give more precise information.
+        eprintln!("{}", mess);
```
piodul, line 23: This only prints a message but does not fail the test in any way.

piodul (changes requested):
A couple of things:

- It might be simpler to not use generics here. With current approach, support for things such as multi-column partition keys might be hard to add. To work around `CqlValue` not being hashable, I'd suggest adding a new enum `PkValue` which implements Eq and Hash but is able to hold only a small subset of possible `CqlValue`s (let's start with 1-2 variants).

- I'd recommend that the `TestConsumer` does not perform any assertions by itself, instead it can just put consumed rows in a `HashMap<Vec<PkValue>, Whatever>`. At the end of the test you can go over every partition key and report inconsistencies, just like `compare_changes` in `replication_tests.rs` does.

As for running this test on CI, it is not true that with one node there is only one stream. By default, the number of streams is equal to `256 * shard_count * node_count`, so even with one node the test makes sense.

scylla-cdc/src/lib.rs
procrastinationfighter: To create the schema without complicating the loop body.
It's only a placeholder though, this function's body will be replaced as soon as the library is finished.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,100 @@ pub mod stream_generations;
...
+    }
+
+    // First one is vec of clustering keys, the second are other columns.
+    // Only not null values should be present in these vectors.
+    // If a value v is deleted, insert pair ("v", None) to the second vector.
+    // [TODO]: Add support for non-frozen collections
+    type Operation<> = (OperationType, Vec<(&'static str, CqlValue)>, Vec<(&'static str, Option<CqlValue>)>);
```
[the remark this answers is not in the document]
procrastinationfighter: No they're not, my bad - I used lifetimes before and forgot to remove the brackets.

procrastinationfighter:
@piodul as for the numbers of streams, how can we force Scylla to use many of them? When I ran the `sample_test`, all entries had the same `stream_id` (there are 100 different int values).

scylla-cdc/src/lib.rs
piodul: Even if the code is going to be eventually replaced, I would like to know the reason. Why getting the first row is needed to create the schema? You already have the `rows_stream` object and the `get_column_specs()` method is supposed to work without before calling `next()` or iterating.

procrastinationfighter: Nevermind then, some kind of mind lag must have occured, you are correct.

piodul:
> @piodul as for the numbers of streams, how can we force Scylla to use many of them? When I ran the `sample_test`, all entries had the same `stream_id` (there are 100 different int values). Does it involve exploring the partitioner?

I don't think there is a need to force it. When I run a fresh Scylla instance it creates a generation with multiple streams, and when I insert several keys each one of them lands in a different stream. Which Scylla version do you use? What is the output of `select count(*) from system_distributed.cdc_streams_descriptions_v2`?

procrastinationfighter:
As of stream ids: false alarm, another overlook of mine. Everything was fine, I just didn't check it correctly in the `cqlsh`.
I applied suggested changes. The tests compares changes for each partition key and stops at the first inconstistency for each row. 

One additional commit was added - I added `Clone` trait for the `OperationType` enum.

Also, Clippy suddenly went mad at some code in the replicator, probably due to an update, because it doesn't show me any errors on my computer.

I will add support for complex partition keys later.

procrastinationfighter:
I added support for complex partition keys. I didn't add support for nested tuple partition keys, but at the end I found on that construction like `primary key (((pk1, pk2), pk3), ck))` are incorrect, so I guess it's okay.
If everything is fine, then the only thing to add to these tests is to use the library's interface in them, but that can't be done before #12 is solved, so I think that's all for now.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+        session.query("CREATE KEYSPACE IF NOT EXISTS e2e_test WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }", ()).await?;
+        session.use_keyspace("e2e_test", false).await?;
+
+        session
+            .query(format!("DROP TABLE IF EXISTS {}", table_name), ())
+            .await?;
+
+        let (create_query, insert_query, update_query) = if pk_type_names.len() == 1 {
+            (
+                format!(
```
piodul, line 161: For the sake of simplicity, let's get rid of the special case for `pk_type_names.len() == 1` and just use the `pk1` name in such case.

piodul (changes requested):

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+                CqlValue::Text(s) => Some(PrimaryKeyValue::Text(s.clone())),
+                _ => None,
+            }
+        }
+
+        pub fn as_cql(&self) -> CqlValue {
+            match self {
+                PrimaryKeyValue::Int(x) => CqlValue::Int(*x),
+                PrimaryKeyValue::Text(s) => CqlValue::Text(s.clone()),
+                PrimaryKeyValue::List(v) => {
```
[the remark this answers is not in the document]
procrastinationfighter: It's related with the other comment. `PrimaryKeyValue::List` was meant to indicate that the primary key is complex. Since it will not, I'll add the other conversion.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+            }
+        }
+
+        test.test_cdc().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn int_string_pk_test() {
+        let session = get_session().await.unwrap();
+        let mut test = Test::new(session, "int_test", vec!["int", "text"])
```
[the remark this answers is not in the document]
procrastinationfighter: No, it wasn't, I'll fix that.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+            let operation = Operation::new(OperationType::RowUpdate, ck, v);
+            self.push_back(pk, operation);
+
+            Ok(())
+        }
+
+        async fn compare(
+            &mut self,
+            result: Arc<Mutex<HashMap<PrimaryKeyValue, VecDeque<Operation>>>>,
+        ) -> bool {
```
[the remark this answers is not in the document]
procrastinationfighter: I'm not sure I understand. Is using `unwrap_or_default` enough (since it will force the tests to write information about fail reason)?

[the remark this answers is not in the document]
procrastinationfighter: Is it still ok for `next_expected` to be unwrapped? The `None` case should never happen, because of the loop condition.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+                    CqlValue::List(v.iter().map(PrimaryKeyValue::as_cql).collect())
+                }
+            }
+        }
+    }
+
+    #[derive(Debug, Eq, PartialEq)]
+    struct Operation {
+        operation_type: OperationType,
+        clustering_key: i32,
```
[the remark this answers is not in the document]
procrastinationfighter: Changing it to `Option` was trivial. As for changing `i32` to `CqlValue`, see my answer for your other comment.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,440 @@ pub mod stream_generations;
...
+        pub fn from_cql(cql_val: &CqlValue) -> Option<PrimaryKeyValue> {
+            match cql_val {
+                CqlValue::Int(x) => Some(PrimaryKeyValue::Int(*x)),
+                CqlValue::Text(s) => Some(PrimaryKeyValue::Text(s.clone())),
+                CqlValue::List(v) => Some(PrimaryKeyValue::List(
+                    // [FIXME]: A monad-like construction would be nice to avoid unwrapping without much code.
+                    v.iter()
+                        .map(|x| PrimaryKeyValue::from_cql(x).unwrap())
+                        .collect(),
+                )),
```
procrastinationfighter, line 53: Is there a nice way to map these lists? My intention was: if any value in the list can't be converted into `PrimaryKeyValue`, return `None`.

piodul: You [can](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3COption%3CA%3E%3E) collect an `Iterator<Item = Option<T>>` to `Option<Vec<T>>`. This should work (I hope, I didn't try compiling it):

```rust
v.iter()
    .map(PrimaryKeyValue::from_cql)
    .collect::<Option<Vec<_>>()?
```

This trick also works with collecting [`Result`s](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3CResult%3CA%2C%20E%3E%3E).

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,426 @@ pub mod stream_generations;
...
+                }
+            }
+        }
+    }
+
+    #[derive(Debug, Eq, PartialEq)]
+    struct Operation {
+        operation_type: OperationType,
+        clustering_key: i32,
+        value: Option<i32>,
```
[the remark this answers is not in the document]
piodul: OK, as we are just concentrating on verifying the order of the operations, we can stick to `i32` for now.

Technically speaking, an end-to-end test by definition is supposed to test all of the features, and adding support for more CQL types and CDC operations will allow us to verify if more complex operations are represented correctly as CDCRow (e.g. operations on non-frozen columns). However, as we have limited time and this should already be covered in part by the replicator tests, we can postpone it to later or never.

scylla-cdc/src/lib.rs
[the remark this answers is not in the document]
piodul: You can change the inner loop to `while let Some(next_expected) = expected_operations.pop_front() { /* ... */ }`.

scylla-cdc/src/lib.rs
```diff
@@ -5,9 +5,440 @@ pub mod stream_generations;
...
+
+    // The driver's CqlValue cannot be used as HashMap key,
+    // because it doesn't have the Eq trait.
+    #[derive(Debug, Eq, PartialEq, Hash)]
+    enum PrimaryKeyValue {
+        // Name consistency with CqlValue from the driver is recommended.
+        Int(i32),
+        Text(String),
+        List(Vec<PrimaryKeyValue>),
+        ComplexPK(Vec<PrimaryKeyValue>),
```
piodul, line 32: Can't we just use `Vec<PrimaryKeyValue>` instead of `PrimaryKeyValue` as the key in `OperationsMap`? This should eliminate the need for `ComplexPK`.

procrastinationfighter: I thought it was more intuitive in the case of using a one-column pk. Maybe you're right, I'll have to see which one looks better.

procrastinationfighter:
I applied the changes and tested the library. I tried to keep the changes readable, but it seems that force-pushing does not keep the commit history, so you can't see the history before squashing.
Long story short: 
- I decided to remove the `ComplexPK` type and use always vectors, like you suggested. 
- The `test_cdc` method now uses the library and accepts a `start` timestamp.

scylla-cdc/src/e2e_tests.rs
```diff
@@ -0,0 +1,463 @@
...
+                start,
+                end,
+                time::Duration::from_millis(WINDOW_SIZE),
+                time::Duration::from_millis(SAFETY_INTERVAL),
+                time::Duration::from_millis(SLEEP_INTERVAL),
+                factory,
+            );
+
+            sleep(time::Duration::from_secs(15)).await;
+            tester.stop();
```
piodul, line 357: This causes the test to depend on timing. You can't be sure that everything completes within 15ms. Moreover, after calling `stop()` the log reader worker will stop ASAP but not immediately.

As you already provided an end timestamp in the constructor, it will stop itself after reaching it. In order to wait until the worker actually finishes, you need to `.await` the handle returned from `CDCLogReader::new`.

piodul (changes requested):
Pointed out a small number of issues that need to be fixed, but otherwise looks good.

scylla-cdc/src/e2e_tests.rs
procrastinationfighter: I'm not sure that works. The tests don't come to an end, I even got messages like this one: `test e2e_tests::tests::e2e_test_int_pk has been running for over 60 seconds`. 
I replaced these two lines with this one: 
```
handle.await.unwrap();
```

piodul: In that case, it sounds like we might have a bug in the log reader and it should be fixed. cc: @Ponewor

piodul: It looks like the CDCLogReader, after creating StreamReaders, should set their upper timestamps to self.end_timestamp, but it doesn't.

procrastinationfighter: That might be it, I just did some trial and error attempts and it seems that the problem is with the default end timestamp - it works fine when we use `stop` or even `stop_at`. I'm going to try it.

procrastinationfighter: Yup, it works.

piodul (approved):

scylla-cdc/src/e2e_tests.rs
```diff
@@ -0,0 +1,461 @@
...
+
+                    if next_expected != next_actual {
+                        eprintln!("Operation no. {} not matching for primary key {:?}.", i, pk);
+                        eprintln!("\tExpected: {:?}, actual: {:?}", next_expected, next_actual);
+                        is_ok = false;
+                        break;
+                    }
+                }
+
+                if is_ok && expected_operations.len() != actual_operations.len() {
```
Ponewor, line 325: This will be ignored if there was a difference in previous loop iterations.

procrastinationfighter: I agree, this was a hotfix because before the message was written every time when something failed earlier. (i.e. we had two messages: some row didn't match and then we did a `break` and another message was written).

PR #51, opened 2022-03-16
Implement range row deletes

### LINKED ISSUE #25: Replicator: support range deletes
https://docs.scylladb.com/using-scylla/cdc/cdc-basic-operations/#row-range-deletions

Ponewor:
The scope of this PR is much smaller than promised due to unexpected technical difficulties. I will try to resolve them later.

Closes #25

### DIFF
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -141,7 +141,7 @@ mod tests {
         ks_src: &str,
         ks_dst: &str,
         name: &str,
-        last_read: &mut u64,
+        last_read: &mut (u64, i32),
     ) -> anyhow::Result<()> {
         let result = session
             .query(
@@ -173,8 +173,9 @@ mod tests {
         for log in result.rows.unwrap_or_default() {
             let cdc_row = CDCRow::from_row(log, &schema);
             let time = cdc_row.time.to_timestamp().unwrap().to_unix_nanos();
-            if time > *last_read {
-                *last_read = time;
+            let batch_seq_no = cdc_row.batch_seq_no;
+            if (time, batch_seq_no) > *last_read {
+                *last_read = (time, batch_seq_no);
                 consumer.consume_cdc(cdc_row).await?;
             }
         }
@@ -409,7 +410,7 @@ mod tests {
         setup_udts(&session, &ks_src, &ks_dst, &udt_schemas).await?;
         setup_tables(&session, &ks_src, &ks_dst, &table_schema).await?;
         session.use_keyspace(&ks_src, false).await?;
-        let mut last_read = 0;
+        let mut last_read = (0, 0);
 
         for operation in operations {
             session.query(operation, []).await?;
@@ -888,7 +889,7 @@ mod tests {
             .unwrap();
 
         session.use_keyspace(&ks_src, false).await.unwrap();
-        let mut last_read = 0;
+        let mut last_read = (0, 0);
 
         for operation in operations {
             session.query(operation, []).await.unwrap();
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -978,4 +978,31 @@ mod tests {
             .await
             .unwrap();
     }
+
+    #[tokio::test]
+    async fn test_range_delete() {
+        let schema = TestTableSchema {
+            name: "RANGE_DELETE".to_string(),
+            partition_key: vec![("pk1", "int"), ("pk2", "int")],
+            clustering_key: vec![("ck1", "int"), ("ck2", "int"), ("ck3", "int")],
+            other_columns: vec![("v", "int")],
+        };
+
+        let operations = std::iter::repeat(0..5).take(3).multi_cartesian_product().map(|x| {
+            format!("INSERT INTO RANGE_DELETE (pk1, pk2, ck1, ck2, ck3, v) VALUES (0, 0, {}, {}, {}, 0)", x[0], x[1], x[2])
+        }).collect::<Vec<_>>();
+
+        let mut operations: Vec<&str> = operations.iter().map(|x| x.as_str()).collect();
+
+        operations.append(&mut vec![
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND ck1 = 0 AND ck2 > -1 AND ck2 < 1",
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND ck1 = 1 AND ck2 < 2",
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND (ck1, ck2) < (3, 3) AND (ck1, ck2, ck3) > (2, 2, 2)",
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND (ck1, ck2, ck3) > (3, 3, 3)",
+        ]);
+
+        test_replication(&get_uri(), schema, operations)
+            .await
+            .unwrap();
+    }
 }
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -507,18 +507,18 @@ impl ReplicatorConsumer {
         &mut self,
         column_name: &str,
         data: &'a CDCRow<'_>,
-        values_for_update: &mut [&'a CqlValue],
+        values_for_update: &mut [Option<&'a CqlValue>],
         timestamp: i64,
         sentinel: CqlValue,
     ) -> anyhow::Result<()> {
-        let value = data.get_value(column_name).as_ref().unwrap_or(&sentinel);
+        let value = Some(data.get_value(column_name).as_ref().unwrap_or(&sentinel));
         // Order of values: ttl, added elements, pk condition values.
 
         let deleted_set = Set(Vec::from(data.get_deleted_elements(column_name)));
         let mut values_for_update = values_for_update.to_vec();
 
         values_for_update[1] = value;
-        values_for_update.insert(2, &deleted_set);
+        values_for_update.insert(2, Some(&deleted_set));
         // New order of values: ttl, added elements, deleted elements, pk condition values.
 
         self.precomputed_queries
@@ -533,8 +533,8 @@ impl ReplicatorConsumer {
         &mut self,
         column_name: &str,
         data: &'a CDCRow<'_>,
-        values_for_update: &mut [&'a CqlValue],
-        values_for_delete: &[&CqlValue],
+        values_for_update: &mut [Option<&'a CqlValue>],
+        values_for_delete: &[Option<&CqlValue>],
         timestamp: i64,
         sentinel: CqlValue,
     ) -> anyhow::Result<()> {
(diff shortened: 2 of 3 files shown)


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -27,6 +27,9 @@ pub(crate) struct ReplicatorConsumer {
 
     // Strings for queries created dynamically:
     keys_cond: String,
+
+    // Stores data for left side range delete while waiting for its right counterpart.
+    left_range_details: (Option<usize>, bool, Option<CqlValue>),
```
piodul, line 32: This data is not enough to reconstruct a range delete. Take into account that queries like the following also exist:

```cql
DELETE FROM ks.t WHERE (ck1, ck2) > (1, 2);
DELETE FROM ks.t WHERE (ck1, ck2) > (1, 2) AND (ck1, ck2) < (9, 10);
DELETE FROM ks.t WHERE (ck1, ck2, ck3, ck4) > (1, 2, 3, 4) AND (ck1, ck2) < (9, 10);
```

You will have to keep all `CqlValue`s from the previous row (or at least up to the first `null`).

piodul (changes requested):
This PR does not support all kinds of range deletes - notably, it is possible to provide a tuple-like restriction, e.g. `WHERE (ck1, ck2) < (1, 2)`. Please add support for them and add more tests.

(I now see that the CDC doc page does not mention such range deletes, nevertheless it would be good to support them and I think by doing so the logic will become simpler).

Additionally, please add `Fixes: #25` line to the PR's description.

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: Moreover, please convert the tuple either into separate fields with clear names, or a struct - also with clear field names.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -522,6 +529,128 @@ impl ReplicatorConsumer {
...
+            .iter()
+            .map(|el| el.map(|x| x.clone()))
+            .collect();
+    }
+
+    async fn delete_row_range_right(
+        &self,
+        data: CDCRow<'_>,
+        right_included: bool,
+    ) -> anyhow::Result<()> {
```
piodul, line 545: Wow... that turned out to be quite complicated. I think the following observation should help you simplify - this:

```cql
ck1 = a AND ck2 = b AND ck3 > c AND ck3 < d
```

...is equivalent to:

```cql
(ck1, ck2, ck3) > (a, b, c) AND (ck1, ck2, ck3) < (a, b, d)
```

...therefore you can completely forgo using the first form and only use tuple-inequalities.

Ponewor: That trick won't work for queries like:
`ck1 = a AND ck2 = b` or `ck1 = a AND ck2 = b AND ck3 > c`

We should also keep in mind that in tables with compact storage queries like
`ck1 = a AND ck2 = null`
are also possible and null values do not work well in tuple-inequalities.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -519,6 +529,128 @@ impl ReplicatorConsumer {
         Ok(())
     }
 
+    fn delete_row_range_left(&mut self, data: CDCRow<'_>, included: bool) {
+        let (_, _, values) = self.get_common_cdc_row_data(&data);
+        self.left_range_included = included;
+        self.left_range_values = values[self.table_schema.partition_key.len()..]
+            .iter()
+            .map(|el| el.map(|x| x.clone()))
+            .collect();
```
piodul, line 538: This seems to be a bit roundabout way to obtain a vector of clustering key values. The `get_common_cdc_row_data` computes several values but you only need one, moreover it creates a vector of pk + ck values, you discard the pks and clone the rest. Why not create a vector of cks manually and avoid clones altogether:

```rust
self.left_range_values = self
    .table_schema
    .clustering_key
    .iter()
    .map(|ck| data.take_value(ck))
    .collect();
```

piodul: You can probably do something like this in `delete_row_range_right`, too. Perhaps it's worth extracting to a method.

piodul: Additionally, consider removing the null values from the vector. For range deletes, the cdc row will contain a prefix of non-null ck values and then only nulls. In this case, adding `.take_while(Option::is_some)` before `.collect()`. should do the trick. This might simplify things.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -519,6 +529,128 @@ impl ReplicatorConsumer {
...
+            }
+            Ordering::Greater => {
+                conditions.push(format!(
+                    "({}) {} ({})",
+                    self.table_schema.clustering_key[first_unequal_position..first_left_null_index]
+                        .iter()
+                        .join(","),
+                    greater_than,
+                    &"?,".repeat(first_left_null_index - first_unequal_position)
+                        [..2 * (first_left_null_index - first_unequal_position) - 1]
```
piodul, line 607: The use of slice here is confusing - I suppose its purpose was to skip the last comma. Why not use join like before? Won't something like this work instead?

```rust
std::iter::repeat("?").take(first_left_null_index - first_unequal_position).join(",")
```

piodul (changes requested):

scylla-cdc-replicator/src/replicator_consumer.rs
Ponewor: Apparently Scylla does not produce any CDC log record for `ck1 = a AND ck2 = null`.

piodul: Not sure. Perhaps @kbr- will know more about the interaction of compact storage with CDC?

As I wrote, I don't think we should care about compact storage tables in the replicator.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -1296,4 +1296,49 @@ mod tests {
...
+                vec![Int(0), Int(0), Int(0), Int(-1), Int(1)],
+            ),
+            (
+                "DELETE FROM RANGE_DELETE WHERE pk1 = ? AND pk2 = ? AND ck1 = ? AND ck2 < ?",
+                vec![Int(0), Int(0), Int(1), Int(2)],
+            ),
+            (
+                "DELETE FROM RANGE_DELETE WHERE pk1 = ? AND pk2 = ? AND (ck1, ck2) < ? AND (ck1, ck2, ck3) > ?",
+                vec![Int(0), Int(0), Tuple(vec![Some(Int(3)), Some(Int(3))]), Tuple(vec![Some(Int(2)), Some(Int(2)), Some(Int(2))])]
+            ),
```
piodul, line 1337: I added the following operation and your code fails:

```rust
(
    "DELETE FROM RANGE_DELETE WHERE pk1 = ? AND pk2 = ? AND (ck1, ck2, ck3) > ?",
    vec![Int(0), Int(0), Tuple(vec![Some(Int(2)), Some(Int(2)), Some(Int(2))])],
),
```

```
---- replication_tests::tests::test_range_delete stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Database returned an error: The submitted query has a syntax error, Error message: line 1:92 no viable alternative at input '('', scylla-cdc-replicator/src/replication_tests.rs:1346:14
```

I verified that this error comes from the replicator.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -1296,4 +1296,49 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_range_delete() {
+        let schema = TestTableSchema {
+            name: "RANGE_DELETE".to_string(),
+            partition_key: vec![("pk1", "int"), ("pk2", "int")],
+            clustering_key: vec![("ck1", "int"), ("ck2", "int"), ("ck3", "int")],
+            other_columns: vec![("v", "int")],
+        };
+
+        let operations = vec![
```
piodul, line 1309: I have a feeling that with such a small number of rows inserted we risk the tests not being accurate. I'd recommend inserting more rows - for example, insert a row for each combination of ck1, ck2, ck3 being in range `1..4` (64 inserts, can be automatically generated and inserted into `operations` in the test).

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -173,7 +173,7 @@ mod tests {
         for log in result.rows.unwrap_or_default() {
             let cdc_row = CDCRow::from_row(log, &schema);
             let time = cdc_row.time.to_timestamp().unwrap().to_unix_nanos();
-            if time > *last_read {
+            if time >= *last_read {
```
piodul, line 176: This fix will also accidentally replay changes from the previous operation.

A CDC log table's key is `("cdc$stream_id", "cdc$time", "cdc$batch_seq_no")`, therefore - assuming the same stream id, which is fine here - you can keep track of the last read (cdc$time, cdc$batch_seq_no) pair instead of just cdc$time and keep the strict inqeuality (`>`).

Ponewor: Right

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -690,6 +696,128 @@ impl ReplicatorConsumer {
...
+
+    fn add_range_condition<'a>(
+        &self,
+        conditions: &mut Vec<String>,
+        query_values: &mut Vec<Option<&'a CqlValue>>,
+        values: &[Option<&'a CqlValue>],
+        first_unequal_position: usize,
+        relation: &str,
+    ) {
+        if values[first_unequal_position].is_some() {
```
piodul, line 787: Is this correct? How does this work e.g. for `ck1 = 1 AND ck2 < 2`? Won't this accidentally skip the lower bound? The correct conditions should be `(ck1) < (1) AND (ck1, ck2) > (1, 2)`.

I added some debug prints to the replicator logic and it seems that it generates a wrong query for the second delete in the `test_range_delete`, but the test accepts it although the incorrect delete should delete more rows than necessary. I wonder what's wrong (with the code or my understanding).

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -978,4 +978,31 @@ mod tests {
...
+        };
+
+        let operations = std::iter::repeat(1..5).take(3).multi_cartesian_product().map(|x| {
+            format!("INSERT INTO RANGE_DELETE (pk1, pk2, ck1, ck2, ck3, v) VALUES (0, 0, {}, {}, {}, 0)", x[0], x[1], x[2])
+        }).collect::<Vec<_>>();
+
+        let mut operations: Vec<&str> = operations.iter().map(|x| x.as_str()).collect();
+
+        operations.append(&mut vec![
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND ck1 = 0 AND ck2 > -1 AND ck2 < 1",
```
piodul, line 998: This delete does not modify the table. No rows are generated for `ck = 0`.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -978,4 +978,31 @@ mod tests {
...
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND ck1 = 1 AND ck2 < 2",
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND (ck1, ck2) < (3, 3) AND (ck1, ck2, ck3) > (2, 2, 2)",
+            "DELETE FROM RANGE_DELETE WHERE pk1 = 0 AND pk2 = 0 AND (ck1, ck2, ck3) > (2, 2, 2)",
```
piodul, line 1001: I know that this is operation is based on the query string I gave you, but it would be better if all deleted ranges in the test were disjoint. So, how about `(ck1, ck2, ck3) > (3, 3, 3)`?

piodul (approved):

PR #53, opened 2022-04-02
Replicator: Precompute dynamic queries

### LINKED ISSUE #33: Replicator: Precompute queries
The replicator creates some queries dynamically ([example](https://github.com/piodul/scylla-cdc-rust/pull/28#discussion_r782377259)). These queries should be precomputed.

Gor027:
Fixes #33

This PR includes a small design refactoring in `ReplicatorConsumer`, in order to not pass the whole `self` object to the functions, instead only the necessary fields of `ReplicationConsumer` will be passed. This will allow us to easily add a new mutable field to the `ReplicatorConsumer` struct and update it in the functions.
The dynamic queries are lazily precomputed and stored in a map, however, a dynamic query to delete UDT elements is strictly dependent on the set of indices of UDT fields to be deleted. So, that part is left unmodified.

### DIFF
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -89,34 +89,54 @@ impl PrecomputedQueries {
         })
     }
 
-    async fn delete_partition(&self, values: &[impl Value], timestamp: i64) -> anyhow::Result<()> {
-        self.run_prepared_statement(self.partition_delete_query.clone(), values, timestamp)
-            .await?;
+    async fn delete_partition(
+        &mut self,
+        values: &[impl Value],
+        timestamp: i64,
+    ) -> anyhow::Result<()> {
+        run_prepared_statement(
+            &self.destination_table_params.dest_session,
+            self.partition_delete_query.clone(),
+            values,
+            timestamp,
+        )
+        .await?;
 
         Ok(())
     }
 
-    async fn insert_value(&self, values: &[impl Value], timestamp: i64) -> anyhow::Result<()> {
-        self.run_prepared_statement(self.insert_query.clone(), values, timestamp)
-            .await?;
+    async fn insert_value(&mut self, values: &[impl Value], timestamp: i64) -> anyhow::Result<()> {
+        run_prepared_statement(
+            &self.destination_table_params.dest_session,
+            self.insert_query.clone(),
+            values,
+            timestamp,
+        )
+        .await?;
 
         Ok(())
     }
 
-    async fn delete_row(&self, values: &[impl Value], timestamp: i64) -> anyhow::Result<()> {
-        self.run_prepared_statement(self.delete_query.clone(), values, timestamp)
-            .await?;
+    async fn delete_row(&mut self, values: &[impl Value], timestamp: i64) -> anyhow::Result<()> {
+        run_prepared_statement(
+            &self.destination_table_params.dest_session,
+            self.delete_query.clone(),
+            values,
+            timestamp,
+        )
+        .await?;
 
         Ok(())
     }
 
     async fn overwrite_value(
-        &self,
+        &mut self,
         values: &[impl Value],
         timestamp: i64,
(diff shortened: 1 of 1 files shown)


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -13,20 +13,30 @@ use scylla::Session;
 
 use scylla_cdc::consumer::*;
 
-pub(crate) struct ReplicatorConsumer {
```
piodul, line 16: The diff of the first commit is very hard to read and review... This looks like a refactor of multiple things at once. Could you break it into multiple commits, each commit representing a single refactoring step so that the commit diffs are easier to read?

Gor027: Actually, a little change in `ReplicatorConsumer` struct causes changes in many places, so a single-refactoring-step commit will also be big if want to make sure that each commit compiles and passes all tests. I will try to break it down into smaller ones.

piodul: It's OK if steps are large, as long as changes they introduce are easy to understand. For example, a step can be:

- Introducing a sub-struct and moving some fields there. If the fields are used in many places, then the diff will be large but most changes will be trivial (changing `self.field` to `self.substruct.field`).
- Converting a method into a stand-alone function,
- Changing a method to precompute prepared queries instead of always sending unprepared queries.

Gor027:
ping @piodul : I would like to make sure that the commits are not messy as before for code review.

piodul:
> ping @piodul : I would like to make sure that the commits are not messy as before for code review.

Thanks, it is much better now.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -297,21 +334,27 @@ async fn update_list<'a>(
...
+            destination_table_params.dest_session.prepare(
+                format!(
+                    "UPDATE {ks}.{tbl} USING TTL ? SET {list}[SCYLLA_TIMEUUID_LIST_INDEX(?)] = ? WHERE {cond}",
+                    ks = destination_table_params.dest_keyspace_name,
+                    tbl = destination_table_params.dest_table_name,
+                    list = column_name,
+                    cond = destination_table_params.keys_cond
+                )
+            ).await?
+        );
```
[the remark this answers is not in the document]
piodul: Moreover, I think that it would make sense to convert the `PrecomputedQueries` struct from being just a data container and promote it into some kind of a cache of prepared statements used by the replicator. It would be responsible for preparing, looking up and executing queries with given parameters, along with formatting the strings (just like the code highlighted above).

It could probably keep the `DestinationTableParams` inside itself, I don't it used outside of that context so IMO it should be fine. By doing this you could also probably avoid converting so many methods into standalone functions and reduce their number of parameters.

piodul (changes requested):

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -135,15 +149,28 @@ impl PrecomputedQueries {
         timestamp: i64,
         column_name: &str,
     ) -> anyhow::Result<()> {
-        run_statement(
+        let overwrite_query = match self.overwrite_queries.get_mut(column_name) {
+            Some(query) => query,
+            None => self
+                .overwrite_queries
+                .entry(column_name.to_string())
+                .or_insert(
```
piodul, line 157: This can be changed to just `.insert(column_name.to_string(), ...)`.

piodul (approved):
Left some comments and suggestions which can be addressed after merge. Otherwise, looks good.

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: Disregard my comment, I realized that you need to obtain a reference to the element... `entry().or_insert()` sounds OK.

PR #54, opened 2022-04-03
Create a minimal library

procrastinationfighter:
Fixes: #12 
This PR transforms the printer's code into the library's interface. 
In order to achieve this, the `StreamReader` was modified and now accepts a `Consumer` instance instead of a callback function.
The Printer was refactored and now uses the created interface.


scylla-cdc/src/cdc_reader.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::consumer::ConsumerFactory;
+use crate::reader::StreamReader;
+use crate::stream_generations::GenerationFetcher;
+
+pub struct CDCReader {
+    // Tells the worker to stop
+    // Usage of the "watch" channel will make it possible to change the the timestamp later,
+    // for example if somebody loses patience and wants to stop now not later
+    end_timestamp: tokio::sync::watch::Sender<Duration>,
+}
```
procrastinationfighter, line 22: I don't like the name `Reader`, because it may suggest that it's something similar to `StreamReader`, but I couldn't come up with a better name - I'm open for suggestions.

scylla-cdc/src/cdc_reader.rs
```diff
@@ -0,0 +1,213 @@
...
+            start_timestamp,
+            end_timestamp,
+            window_size,
+            safety_interval,
+            sleep_interval,
+            end_timestamp_receiver,
+            consumer_factory,
+        );
+        let (_, handle) = tokio::task::spawn(async move { worker.run().await }).remote_handle();
+        let printer = CDCReader {
```
procrastinationfighter, line 53: This piece of code (spawning) required the `ConsumerFactory` object to implement also a `Send` trait. From what I've seen, it's probably automatically generated if all members of a structure implement Send, so it's probably not a bothering issue, is it?

scylla-cdc/src/cdc_reader.rs
```diff
@@ -0,0 +1,213 @@
...
+                            let keyspace = self.keyspace.clone();
+                            let table_name = self.table_name.clone();
+                            let consumer = self.consumer_factory.new_consumer();
+                            tokio::spawn(async move {
+                                reader
+                                    .fetch_cdc_with_consumer(keyspace, table_name, consumer)
+                                    .await
+                            })
+                        })
+                        .collect();
```
procrastinationfighter, line 199: I didn't manage to create a new consumer asynchronously, so I removed the `async` trait from `ConsumerFactory`. An alternative is to use [block_on](https://docs.rs/futures/0.2.0/futures/executor/fn.block_on.html), but I think that in this way we show the user that it doesn't matter if the consumers can be created asynchronously.

procrastinationfighter: Also, the spawning forced me to require consumers to implement `Send`. Just like with `ConsumerFactory` (see my other comment).

scylla-cdc/src/reader.rs
```diff
@@ -101,6 +102,78 @@ impl StreamReader {
...
+        &self,
+        keyspace: String,
+        table_name: String,
+        mut consumer: Box<dyn Consumer + Send>,
+    ) -> anyhow::Result<()> {
+        // [TODO]: This function is almost identical to fetch_cdc,
+        //      but it uses the Consumer trait as a callback.
+        //      The reason for leaving the old function are the tests,
+        //      there is no obvious solution as for what to do with them.
+
```
procrastinationfighter, line 116: This temporary solution is left, because I am not sure what to do with the old tests - if the old `fetch_cdc` function is removed, we will have to fix them.

scylla-cdc-printer/src/printer.rs
```diff
@@ -1,220 +1,44 @@
...
+impl Consumer for PrinterConsumer {
+    async fn consume_cdc(&mut self, data: CDCRow<'_>) -> anyhow::Result<()> {
+        // [TODO]: Add prettier printing
+        println!("time: {}, batch_seq_no: {}, end_of_batch: {}, operation: {:?}, ttl: {:?}",
+                 data.time,
+                 data.batch_seq_no,
+                 data.end_of_batch,
+                 data.operation,
+                 data.ttl);
+        Ok(())
```
procrastinationfighter, line 17: I suggest leave the pretty printing for another PR.

scylla-cdc/src/cdc_reader.rs
piodul: It reads a whole CDC log, so maybe `CDCLogReader`?

scylla-cdc/src/cdc_reader.rs
piodul: Yes, requiring `Send` is fine. However, I think that `ConsumerFactory` should require `Sync`, which is a stronger trait than `Send`. This will allow the factory to be shared by multiple tasks and you will be able to keep `ConsumerFactory::new_consumer` async and make it possible to call it concurrently.

scylla-cdc/src/reader.rs
piodul: Won't just passing a lambda to `fetch_cdc` which calls the `consumer.consume_cdc` work? Something like that (in `CDCReaderWorker::run`):

```rust
let keyspace = self.keyspace.clone();
let table_name = self.table_name.clone();
let mut consumer = self.consumer_factory.new_consumer();
tokio::spawn(async move {
    reader
        .fetch_cdc(keyspace, table_name, |row| consumer.consume_cdc(row))
        .await
})
```

piodul: The `fetch_cdc` will have to be modified so that the lambda/closure is able to return `anyhow::Result<()>`.

scylla-cdc/src/cdc_reader.rs
piodul: Consumers can be `Send`, it's practically required when you work with `tokio` and want an object alive across `.await` points because `tokio`'s scheduler is allowed to move tasks between worker threads.

I'd recommend putting the `Sync` and `Send` requirements in the traits' definitions in order to simplify things:

```rust
#[async_trait]
pub trait Consumer: Send {
    /* ... */
}

#[async_trait]
pub trait ConsumerFactory: Sync {
    /* ... */
}
```

procrastinationfighter:
Ok, but we need to transform the row into a `CDCRow`, and to do it we need
the schema, and to create the schema, we need to run at least one query to
get the specs. I'm not sure how to solve this issue without doing a dummy
query in the `run` function.
(I'm sorry if this message appears in some weird place, I'm using the email
interface for the first time)

Piotr Dulikowski ***@***.***> schrieb am Mo., 4. Apr. 2022,
19:52:

piodul:
> Ok, but we need to transform the row into a `CDCRow`, and to do it we need the schema, and to create the schema, we need to run at least one query to get the specs. I'm not sure how to solve this issue without doing a dummy query in the `run` function. (I'm sorry if this message appears in some weird place, I'm using the email interface for the first time)

OK, I see the problem now. Let's start with adapting the existing `fetch_cdc` function and the tests so that the callback takes a `CDCRow` instead of `Row`, then rebase the code from the PR on top of that.

You will probably have to modify the `fetch_callback` quite a bit - you won't be able to push the CDCRow into an outer vector because of lifetimes - so you will probably have to extract values from the CDC row and convert it into a tuple, struct or something else before putting it into the vector.

scylla-cdc/src/log_reader.rs
```diff
@@ -0,0 +1,213 @@
+use std::sync::Arc;
+
+use anyhow;
+use chrono::Duration;
+use core::time;
```
piodul, line 5: Importing from `core` is a bit strange. This should be the same as `use std::time;`.

scylla-cdc/src/log_reader.rs
```diff
@@ -0,0 +1,213 @@
+use std::sync::Arc;
+
+use anyhow;
+use chrono::Duration;
```
piodul, line 4: Let's import the `std::time::Duration` directly and use qualified `chrono::Duration`. I think it is more common for an unqualified `Duration` to come from `std`, not `chrono`, so this should be less confusing.

scylla-cdc/src/log_reader.rs
```diff
@@ -0,0 +1,213 @@
...
+    safety_interval: Duration,
+    readers: Vec<Arc<StreamReader>>,
+    end_timestamp_receiver: tokio::sync::watch::Receiver<Duration>,
+    consumer_factory: Arc<dyn ConsumerFactory>,
+}
+
+impl CDCReaderWorker {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        session: &Arc<Session>,
```
piodul, line 88: Nitpick: it's better to accept by value if you are going to clone anyway. Now, there is a mandatory clone of the `Arc`, which involves a relatively expensive atomic operations, but if you accept by value then the caller will be able to choose if they want to move the `Arc` in or clone it.

scylla-cdc/src/log_reader.rs
```diff
@@ -0,0 +1,213 @@
...
+                    // that could have been missed because of earlier fetching failures.
+                    // More on this here https://github.com/piodul/scylla-cdc-rust/pull/10#discussion_r826865162
+                    return Ok(());
+                }
+            }
+        }
+    }
+
+    async fn set_upper_timestamp(readers: &[Arc<StreamReader>], new_upper_timestamp: Duration) {
+        // This function does not take &self, because that would require the CDCWorker to be Send.
```
piodul, line 208: Do you know why `CDCReaderWorker` is not `Send`? The printer worker had a similar function and it worked fine.

scylla-cdc-printer/src/printer.rs
```diff
@@ -1,220 +1,42 @@
...
-    table_name: String,
-    start_timestamp: Duration,
-    end_timestamp: Duration,
-    sleep_interval: time::Duration,
-    window_size: Duration,
-    safety_interval: Duration,
-    readers: Vec<Arc<StreamReader>>,
-    end_timestamp_receiver: tokio::sync::watch::Receiver<Duration>,
-}
+struct PrinterConsumerFactory {}
```
piodul, line 19: Nit: you can avoid the brackets after the struct name if it has no fields. This is called a "unit" struct. IMO it looks a bit nicer when you don't have to include the brackets every time you instantiate such struct.

scylla-cdc-printer/src/printer.rs
```diff
@@ -1,220 +1,42 @@
...
-            start_timestamp,
-            end_timestamp,
-            window_size,
-            safety_interval,
-            sleep_interval,
-            end_timestamp_receiver,
+use async_trait::async_trait;
+use scylla_cdc::consumer::{CDCRow, Consumer, ConsumerFactory};
+
+struct PrinterConsumer {}
```
piodul, line 5: Same tip as for `PrinterConsumerFactory`.

scylla-cdc-printer/src/printer.rs
```diff
@@ -1,220 +1,42 @@
...
+impl ConsumerFactory for PrinterConsumerFactory {
+    async fn new_consumer(&self) -> Box<dyn Consumer> {
+        Box::new(PrinterConsumer {})
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use chrono::Duration;
+    use core::time;
```
piodul, line 31: `use std::time;`

piodul (commented):
Left several small nitpicks.

In one commit, you copy the printer and immediately adapt it into a library, and in the next commit you delete most of the code of the printer. Because of that it's hard to verify by looking at the diffs what modifications have been made to the printer. Please try making the modifications more clear, e.g. first refactor the CDCLogPrinter into CDCLogReader but keep both in the same file, then move the CDCLogReader into a separate file.

Apart from that, I think it looks good.

scylla-cdc/src/log_reader.rs
procrastinationfighter: It was probably a relict from the times when `ConsumerFactory` wasn't `Sync` and I forgot to change it, my bad.

procrastinationfighter:
I'm not sure if I modified the history the way you wished, but now when I look into the `Commits` tab on Github, it looks better.
Also, a reminder: these changes have not been tested correctly yet, so some bugs might appear after using the tests from #46.

piodul:
> I'm not sure if I modified the history the way you wished, but now when I look into the `Commits` tab on Github, it looks better. Also, a reminder: these changes have not been tested correctly yet, so some bugs might appear after using the tests from #46.

Yes, that's more or less what I requested, thanks. Let's fix the bugs that will appear along with the end-to-end test pull request. Some bugs may also appear after we run the `scylla-test`.

piodul (approved):

PR #55, opened 2022-04-06
Fix CDCLogPrinter

Ponewor:
During @procrastinationfighter's work on #12 in #54 some serious problems with CDCLogPrinter were revealed.

This PR attempts to fix them.

### DIFF
--- a/scylla-cdc-printer/src/printer.rs
+++ b/scylla-cdc-printer/src/printer.rs
@@ -179,8 +179,6 @@ impl CDCLogPrinterWorker {
                         })
                         .collect();
 
-                    self.set_upper_timestamp(generation.timestamp).await;
-
                     stream_reader_tasks = self
                         .readers
                         .iter()
--- a/scylla-cdc-printer/src/printer.rs
+++ b/scylla-cdc-printer/src/printer.rs
@@ -8,7 +8,6 @@ use futures::stream::{FusedStream, FuturesUnordered, StreamExt};
 use futures::FutureExt;
 use scylla::frame::response::result::Row;
 use scylla::Session;
-use tokio::task::JoinError;
 
 use scylla_cdc::cdc_types::GenerationTimestamp;
 use scylla_cdc::reader::StreamReader;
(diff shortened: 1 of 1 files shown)


scylla-cdc-printer/src/printer.rs
```diff
@@ -298,7 +297,7 @@ mod tests {
             .await
             .unwrap();
 
-        let (mut cdc_log_printer, _handle) = CDCLogPrinter::new(
```
piodul, line 302: A single underscore (`_`) is different from a name starting with an underscore - it means "don't bind at all". In turn, lifetime of the value assigned to the underscore is dropped immediately. This matters here as dropping the `_handle` is supposed to abort the internal CDCLogPrinter task when it is dropped.

Source: https://users.rust-lang.org/t/do-variables-shorten-the-scope-lifetime/5754

piodul: Oh, and I just noticed that the `_handle` doesn't do anything. We have something like this in `CDCLogPrinter::new()`:

```rust
let (_, handle) = tokio::task::spawn(async move { worker.run().await }).remote_handle();
```

[`.remote_handle()`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.remote_handle) returns a pair `(Remote, RemoteHandle)`. The `Remote` is a future which polls the converted-from future and sends the result to the `RemoteHandle`. If `RemoteHandle` is dropped, then `Remote` is aborted and stops immediately.

What you did here was calling `.remote_handle()` on _JoinHandle_. This doesn't do anything useful - dropping `RemoteHandle` will abort `Remote` and the `JoinHandle` will be dropped - but tokio doesn't abort the task when `JoinHandle` is dropped, so nothing happens. Actually, you are immediately dropping `Remote`, so `JoinHandle` is never polled by anybody and is also immediately dropped...

This is how it should look like:

```rust
let (fut, handle) = async move { worker.run().await };
tokio::task::spawn(fut);
```

The tokio task will be responsible for polling the `Remote` future. When `RemoteHandle` is dropped, the `Remote` will be aborted and will resolve immediately. `JoinHandle` is not used and is immediately dropped (the task keeps being run by tokio).

piodul: Right, I forgot about adding a call to `remote_handle()`, sorry - edited the comment.

Here is an example from existing code which works for sure: https://github.com/scylladb/scylla-rust-driver/blob/9446cc9730190143c6a3b361ae7a3f3b314b64da/scylla/src/transport/cluster.rs#L116-L117 .

piodul (approved):

PR #56, opened 2022-04-06
Optimize: Attempt to narrow down time window for StreamReaders in CDCLogPrinter

Ponewor:
As I understand there are no CDC rows with cdc$time less than generation they belong to.

So passing to StreamReader a start_timestamp much earlier than the generation given by stream_ids produces multiple unnecessary queries for rows from empty time intervals.

Am I right?


piodul (approved):
Yes, this makes sense.

PR #62, opened 2022-04-14
Check if replication went correctly after each query

### LINKED ISSUE #60: Modify replication tests so that they check the table state after every query
Currently, tests in `replication_tests.rs` perform a number of queries to the first table, replicate data from the CDC log to the second table and finally compare the contents of both tables, expecting both of them to be equal.

We could improve the tests so that instead of comparing the tables only once at the end, we could compare after each step. So, for each query:

1. Perform the query
2. Replicate the rows which have just appeared
3. Compare the contents of the tables

This will allow the existing tests to catch more bugs - for example, it could happen that query B which appears later overwrites the results of query A, but the query A alone is not replicated correctly.

Credit for the idea: https://github.com/piodul/scylla-cdc-rust/pull/34#discussion_r791618418

Ponewor:
Closes #60

### DIFF
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -136,24 +136,12 @@ mod tests {
         Ok(())
     }
 
-    async fn execute_queries(
-        session: &Session,
-        ks_src: &str,
-        operations: Vec<TestOperation<'_>>,
-    ) -> anyhow::Result<()> {
-        session.use_keyspace(ks_src, false).await?;
-        for operation in operations {
-            session.query(operation, []).await?;
-        }
-
-        Ok(())
-    }
-
     async fn replicate(
         session: &Arc<Session>,
         ks_src: &str,
         ks_dst: &str,
         name: &str,
+        last_read: &mut u64,
     ) -> anyhow::Result<()> {
         let result = session
             .query(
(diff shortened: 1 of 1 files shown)


scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -395,10 +383,13 @@ mod tests {
...
         setup_udts(&session, &udt_schemas).await?;
         setup_tables(&session, &table_schema).await?;
-        execute_queries(&session, operations).await?;
-        replicate(&session, &table_schema.name).await?;
-        compare_changes(&session, &table_schema.name).await?;
-        compare_timestamps(&session, &table_schema).await?;
+        session.use_keyspace("test_src", false).await?;
+        for operation in operations {
+            session.query(operation.0, operation.1).await?;
+            replicate(&session, &table_schema.name).await?;
```
piodul, line 389: It's a bit awkward because `replicate` replays all changes from the CDC log, including those caused by the previous operations. Please modify the `replicate` function so that it keeps track of the last read `cdc$time` and only processes rows with `cdc$time` > the last seen timestamp.

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -1127,11 +1118,15 @@ mod tests {
...
         );
         setup_tables(&session, &schema).await.unwrap();
-        execute_queries(&session, operations).await.unwrap();
 
-        replicate(&session, &schema.name).await.unwrap();
+        session.use_keyspace("test_src", false).await.unwrap();
+
+        for operation in operations {
+            session.query(operation.0, operation.1).await.unwrap();
+            replicate(&session, &schema.name).await.unwrap();
```
piodul, line 1126: Here, too.

piodul (changes requested):
Looks good, apart from a minor comment.

scylla-cdc-replicator/src/replication_tests.rs
Ponewor: As you see, we call `replicate` in at least two places. If we store somehow the last read `cdc$time` won't we get some race conditions after we parallelize the tests?

piodul (approved):

PR #64, opened 2022-04-15
Fix clippy complaints about vectors instead of slices.

### LINKED ISSUE #65: Clippy complains about Vectors
As can be observed [here](https://github.com/piodul/scylla-cdc-rust/runs/6036785446?check_suite_focus=true), some update in clippy breaks our builds. Passing `&mut Vec<_>` seems no longer valid.

Ponewor:
This is some quick and probably dirty (as it introduces some new cloning) fix.
Closes #65

### DIFF
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -140,7 +140,7 @@ impl ReplicatorConsumer {
         &self,
         column_name: &str,
         data: &'a CDCRow<'_>,
-        values_for_update: &mut Vec<&'a CqlValue>,
+        values_for_update: &mut [&'a CqlValue],
         timestamp: i64,
         sentinel: CqlValue,
     ) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -324,6 +324,7 @@ impl ReplicatorConsumer {
                 })
                 .join(",");
 
+            let values_for_update = &mut values_for_update.to_vec();
             values_for_update.remove(1);
```
piodul, line 328: Tip: `run_statement` accepts an immutable slice, so no need to physically pop the value, you can just shorten the slice and no cloning will be needed:

```rust
let values_for_update = &values_for_update[..values_for_update.len()-1];
```

piodul: Oh, wait, `.remove(1)` removes the first element, not one element from the end. It's OK, then - my bad.

piodul (approved):

PR #66, opened 2022-04-15
Cleanup datetimes

Ponewor:
I am a bit confused by the types we're using to handle datetimes so this is one step to improve them


scylla-cdc/src/cdc_types.rs
```diff
@@ -40,3 +40,13 @@ impl StreamID {
         StreamID { id: stream_id }
     }
 }
+
+pub trait ToTimestamp {
```
piodul, line 44: I don't see it being used in any public interface, so let's keep it `pub(crate)` for now so that users of the library won't depend on it.

Ponewor: Well, printer needs it to be public.

piodul: That's not a good reason to keep it public. This is a helper trait for internal purposes of both the library and the replicator, and the library's interface does not need it. We don't want to support it.

While not ideal, I'd recommend just introducing a duplicated definition of the trait in the replicator.

piodul (approved):

PR #67, opened 2022-04-15
Parallelize tests

### LINKED ISSUE #59: Allow running multiple tests in parallel
The tests should be refactored, because currently they can be run only on one thread (see `.github/workflows/rust.yml`) and reuse keyspaces and tables, which can result in some bugs (like [this one](https://github.com/piodul/scylla-cdc-rust/pull/54/commits/d4f3cb46677dae697780c2bf0c9704f7bd7ebe6c)). This will also enable the e2e tests to run on many threads, which might uncover some bugs. 
Similar issue in the driver: https://github.com/scylladb/scylla-rust-driver/pull/388
Luckily, in our case, the keyspace names are already injected into queries, but we happen to name them in the same way in different tests (eg. "Test"). We also reuse some tables and table names across the tests, that could also be modified to be on the safe side.

It also would be nice if the tests cleaned up after a success, but not necessary.

Ponewor:
Resolves #59

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,5 +29,4 @@ jobs:
       - name: Build
         run: cargo build --verbose
       - name: Run tests
-        # test threads must be one because else database tests will run in parallel and will result in flaky tests
-        run: cargo test --verbose -- --test-threads=1
+        run: cargo test --verbose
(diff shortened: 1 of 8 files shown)


piodul (changes requested):
Looks good, only one request from me.

piodul:
Argh, I didn't know that GitHub's "Resolve conflicts" UI works by creating merging `main` into current branch...

@Ponewor please rebase and resolve conflicts, then I'll merge.

piodul (approved):

Issue #57, opened 2022-04-06
Saving progress


piodul:
Although a client application is supposed to continuously fetch and process data from the CDC log, it may be stopped either due to user's wish or some unforseen environmental factor (e.g. there is a power outage). It would be useful for the application to be able to restart and continue from the place where it previously ended processing.

The task is to design and implement possibility of saving progress in the library.

Requirements:

- The library should offer an interface and let the users implement their own methods of saving progress.
- The library should provide an implementation which saves progress in Scylla, in a user-chosen table.
- It's not necessary for the system to be perfect, i.e. it's fine if the library reads some rows again after restart, so saving progress periodically is fine. On the other hand, it __musn't__ skip rows.

Some notes/tips:

- The interface methods should be asynchronous and should be able to return errors.
- Each stream reader should periodically save progress, independently of each other. The frequency of saving progress should be configurable.
- Even if a stream reader reads several empty windows, this still counts as progress and should be saved. Otherwise, after restoring progress, the stream reader will have to go through those windows again and it will take a long time.

piodul:
@macher259

PR #69, opened 2022-04-20
Saving progress

macher259:
Fixes: #57 

Description: 
Although a client application is supposed to continuously fetch and process data from the CDC log, it may be stopped either due to user's wish or some unforeseen environmental factor (e.g. there is a power outage). It would be useful for the application to be able to restart and continue from the place where it previously ended processing.

Created new trait responsible for creating checkpoints along with default implementation.  Also created a function spawning new tokio task solely responsible for periodically creating these checkpoints. By using [RemoteHandle](https://rust-lang-nursery.github.io/futures-api-docs/0.1.27/futures/future/struct.RemoteHandle.html) we should be able to manage said task.

W.I.P.: Need to create more tests, integrate it into existing stream readers and implement reading checkpoints.


scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,192 @@
...
+    // Creates new table with specified schema in the given keyspace with provided name.
+    // If a table with such name already exists, doesn't do anything.
+    async fn create_table(
+        session: &Session,
+        keyspace: &str,
+        checkpoint_tablename: &str,
+    ) -> anyhow::Result<()> {
+        let schema = format!(
+            "CREATE TABLE IF NOT EXISTS {}.{} (
+            time timestamp PRIMARY KEY,
```
piodul, line 106: I'm not sure if this is meant to be the final schema, but in its current state it's not going to work well. After a long time, you will have lots of timestamps in the table. Partition keys are not sorted by value, so on restart you will have to fetch all of them and compute the maximum manually.

Moreover, we would like to be able to save progress on a per-stream basis. Creating a separate table for each stream isn't really feasible, instead you should keep the last timestamp for each stream in a separate partition.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,192 @@
...
+    tokio::task::spawn(fut);
+
+    handle
+}
+
+/// Customizable trait responsible for creating checkpoints.
+#[async_trait]
+pub trait CDCCheckpointMaker: Send + Sync {
+    /// Creates new checkpoint with given timestamp.
+    async fn make_checkpoint(&mut self, timestamp: Duration) -> anyhow::Result<()>;
```
piodul, line 53: Let's change this method to take `&self`. The user will probably provide a single instance of the maker and it will be shared by multiple consumers, and the user will be responsible for synchronization.

Alternatively, you can consider the factory pattern, just like CDCConsumerFactory + CDCConsumer. Then, you can keep the `&mut self`

scylla-cdc/src/checkpoints.rs
piodul: Information about the current generation should also be kept, so that - on start - the log reader knows which streams to read from and which streams' last read timestamps to ask for.

piodul (commented):
Left some preliminary comments.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,192 @@
...
+    async fn test_creating_single_checkpoints() {
+        let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+
+        let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+        let ks = create_test_keyspace(&session).await;
+        let big_duration = Duration::from_secs(1234567); // Some big duration that will not pass during the test execution.
+        let table_name = "single_checkpoint";
+        let (_, rx) = watch::channel(big_duration);
+        let session = Arc::new(session);
+        let cp_maker =
```
Ponewor, line 174: Mr. Macher, NOOOOO!!!

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    tokio::task::spawn(fut);
+
+    handle
+}
+
+/// Customizable trait responsible for creating checkpoints.
+#[async_trait]
+pub trait CDCCheckpointMaker: Send + Sync {
+    /// Creates new checkpoint with given timestamp.
+    async fn make_checkpoint(&self, checkpoint: Checkpoint) -> anyhow::Result<()>;
```
Ponewor, line 52: There is something wrong with the naming here and it manifests itself across the whole module. If `make_checkpoint` creates a checkpoint then why does it receive `checkpoint: Checkpoint` as an argument?

Either `Checkpoint` is not actually a checkpoint and has to be made/created/saved yet and you should rename it properly or `CDCCheckpointMaker` and corresponding methods should actually be renamed to something like `CDCCheckpointSaver` or maybe `CDCCheckpointManager`.

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
Ponewor: Also, what given timestamp do you mean in that docstring?

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+use tokio::time::sleep;
+
+/// Alias representing timestamp.
+/// (Last read timestamp, one of stream ids tracked by stream reader, generation of stream)
+pub type Checkpoint = (Duration, StreamID, GenerationTimestamp);
+
+/// Starts creating checkpoints starting from now.
+/// Spawns a new task that periodically creates new checkpoint.
+/// Receives value of the next time window via channel.
+/// We are using tokio::sync::watch channel, because we only need the latest value.
```
Ponewor, line 20: -> 
```rust
We are using [`tokio::sync::watch`][`tokio::sync::watch::channel`] channel, because we only need the latest value.
```

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// Loads last seen generation from the checkpoint table.
+    async fn load_last_generation(&self) -> anyhow::Result<Option<GenerationTimestamp>>;
+    /// Loads last seen timestamp for given stream id from the checkpoint table.
+    async fn load_last_checkpoint(
+        &self,
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>>;
+}
+
+/// Default implementation for CDCCheckpointMaker trait.
```
Ponewor, line 62: ->
```rust
/// Default implementation for [`CDCCheckpointMaker`] trait.
```

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+pub struct CDCDefaultCheckpointMaker {
+    session: Arc<Session>,
+    keyspace: String,
+    checkpoint_tablename: String,
+    prepared_stmt: PreparedStatement,
+}
+
+impl CDCDefaultCheckpointMaker {
+    /// Creates new CDCDefaultCheckpointMaker.
```
Ponewor, line 71: ->
```rust
[`CDCDefaultCheckpointMaker`]
```

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// If given None value as table name,
```
Ponewor, line 72: -> `None`

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// it uses default value "cdc_checkpoint".
```
Ponewor, line 73: The value you've described here is different from the one you're actually using.

Or maybe I am wrong and Scylla doesn't distinguish between upper and lower case but it is still confusing to use both ways here.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+            checkpoint_tablename,
+            prepared_stmt,
+        };
+
+        Ok(cp_maker)
+    }
+
+    // Creates new table with specified schema in the given keyspace with provided name.
+    // If a table with such name already exists, doesn't do anything.
+    async fn create_table(
```
Ponewor, line 105: Maybe give it more specific name or at least make your comment more specific?
At this point this makes me think that this method can be used to create any table I specify.

It would suffice to change the comment to:
```rust
// Creates new checkpoints table in the given keyspace with provided name.
```

or change the method name to `create_checkpoints_table`.

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+                    "SELECT time FROM {}.{}\
+                    WHERE stream_id = (?)\
+                    LIMIT 1",
+                    self.keyspace, self.checkpoint_tablename
+                ),
+                (stream_id,),
+            )
+            .await?
+            .maybe_first_row_typed::<(chrono::Duration,)>()?
+            .map(|time| -> chrono::Duration { time.0 });
```
Ponewor, line 187: Why not just `.map(|time| time.0))`?

Privately, I would be tempted to just write

```rust
Ok(self
    .session
    .query(
        format!(
            "SELECT time FROM {}.{}\
            WHERE stream_id = (?)\
            LIMIT 1",
            self.keyspace, self.checkpoint_tablename
        ),
        (stream_id,),
    )
    .await?
    .maybe_first_row_typed::<(chrono::Duration,)>()?
    .map(|time| time.0))
```

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>> {
+        let timestamp = self
+            .session
+            .query(
+                format!(
+                    "SELECT time FROM {}.{}\
+                    WHERE stream_id = (?)\
+                    LIMIT 1",
```
Ponewor, line 180: Are those slashes required?

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        let checkpoint_tablename = tablename.unwrap_or_else(|| "CDC_CHECKPOINT".to_string());
+
+        CDCDefaultCheckpointMaker::create_table(&session, &keyspace, &checkpoint_tablename).await?;
+        session.await_schema_agreement().await?;
+
+        let prepared_stmt = session
+            .prepare(format!(
+                "UPDATE {}.{} SET time = (?)
+                WHERE stream_id = (?)
+                AND generation = (?)",
```
Ponewor, line 89: Do we need parenthesis around the question marks?

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    async fn load_last_checkpoint(
+        &self,
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>> {
+        let timestamp = self
+            .session
+            .query(
+                format!(
+                    "SELECT time FROM {}.{}\
+                    WHERE stream_id = (?)\
```
Ponewor, line 179: Do we need parenthesis around the question marks?

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>>;
+}
+
+/// Default implementation for CDCCheckpointMaker trait.
+pub struct CDCDefaultCheckpointMaker {
+    session: Arc<Session>,
+    keyspace: String,
+    checkpoint_tablename: String,
```
Ponewor, line 66: Notice that whenever you're using one of those fields you're using the other one as well and it's always something like
```rust
format!("{}.{}", self.keyspace, self.checkpoint_tablename)
```

Maybe you could just join them into one field?

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+            prepared_stmt: self.prepared_stmt.clone(),
+        }
+    }
+}
+
+#[async_trait]
+impl CDCCheckpointMaker for CDCDefaultCheckpointMaker {
+    /// Writes new record containing given timestamp to the checkpoint table.
+    async fn make_checkpoint(&self, checkpoint: Checkpoint) -> anyhow::Result<()> {
+        let timestamp = Timestamp(chrono::Duration::from_std(checkpoint.0).unwrap());
```
Ponewor, line 141: Do not use `unwrap` unnecessarily, here you can just use `?`.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>>;
+}
+
+/// Default implementation for CDCCheckpointMaker trait.
+pub struct CDCDefaultCheckpointMaker {
+    session: Arc<Session>,
+    keyspace: String,
+    checkpoint_tablename: String,
+    prepared_stmt: PreparedStatement,
```
Ponewor, line 67: Can we have a better name for this field? `update_stmt` seems a decent choice.

scylla-cdc/src/checkpoints.rs
Ponewor: Also ->
```
`"cdc_checkpoint"`
```

scylla-cdc/src/checkpoints.rs
Ponewor: Or maybe `make_checkpoint_stmt` or shorter `make_checkpoint` depending on how you'll rename other entities.

Ponewor (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        keyspace: &str,
+        checkpoint_tablename: &str,
+    ) -> anyhow::Result<()> {
+        let schema = format!(
+            "CREATE TABLE IF NOT EXISTS {}.{} (
+            stream_id blob PRIMARY KEY,
+            time timestamp,
+            generation timestamp,
+            )
+            WITH CLUSTERING ORDER BY (generation DESC)",
```
piodul, line 116: Setting "generation" as the clustering key does not make it possible to find the maximum generation with `SELECT generation FROM x LIMIT 1`.

Performing an unrestricted SELECT returns the contents of a table ordered by:

1. Partition key token; the token is a hash of the partition key, so the order looks random,
2. Clustering key, as defined by CLUSTERING ORDER.

Because of that, the rows returned by `SELECT generation FROM x` won't be necessarily ordered by generation. Besides, such queries are quite heavy (and usually slow) because they involve all nodes in the cluster, even if you use `LIMIT 1`.

What I would recommend is to have a special, single partition which holds the current generation and read/update it when needed. You can, for example, use an empty string as `stream_id` to identify the special partition.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+impl Clone for CDCDefaultCheckpointMaker {
+    fn clone(&self) -> Self {
+        CDCDefaultCheckpointMaker {
+            session: self.session.clone(),
+            keyspace: self.keyspace.clone(),
+            checkpoint_tablename: self.checkpoint_tablename.clone(),
+            prepared_stmt: self.prepared_stmt.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
piodul: It seems so, Moreover, I don't understand why `Clone` is needed?

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /// Loads last seen generation timestamp or 0 if none.
+    async fn load_last_generation(&self) -> anyhow::Result<Option<GenerationTimestamp>> {
+        let generation = self
+            .session
+            .query(
+                format!(
+                    "SELECT generation FROM {}.{}
+                    LIMIT 1",
```
piodul, line 159: This won't always return the last generation, as I explained in another comment (in `CDCDefaultCheckpointMaker::new`)

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+        Ok(cp_maker)
+    }
+
+    // Creates new table with specified schema in the given keyspace with provided name.
+    // If a table with such name already exists, doesn't do anything.
+    async fn create_table(
+        session: &Session,
+        keyspace: &str,
+        checkpoint_tablename: &str,
+    ) -> anyhow::Result<()> {
```
piodul, line 109: Why not convert it into a method? It already uses most of the `CDCDefaultCheckpointMaker`'s fields.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,256 @@
...
+    let (fut, handle) = async move {
+        loop {
+            let mut checkpoint = receiver.borrow().clone();
+
+            // We are given timestamp of the beginning of the next read window,
+            // so we have to create a checkpoint with lower timestamp.
+            if checkpoint.0 > Duration::from_millis(1) {
+                checkpoint.0 -= Duration::from_millis(1);
+                checkpoint_maker.make_checkpoint(checkpoint).await?;
+            }
```
piodul, line 36: Despite the comment, I still don't understand why this is needed. Could you explain why we need to subtract one millisecond? Why can't we just store the timestamp just at it is?

scylla-cdc/src/log_reader.rs
```diff
@@ -175,6 +231,10 @@ impl CDCReaderWorker {
                                 self.window_size,
                                 self.safety_interval,
                                 self.sleep_interval,
+                                self.should_save_progress,
+                                self.should_resume_progress,
+                                time::Duration::from_millis(3),
```
piodul, line 236: This shouldn't be hardcoded (also, it's definitely too low)

macher259:
Resolved conflicts, applied suggestions and created tests.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,434 @@
...
+use futures::future::RemoteHandle;
+use futures::FutureExt;
+
+use scylla::frame::response::result::CqlValue::Timestamp;
+use scylla::prepared_statement::PreparedStatement;
+use scylla::Session;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::RwLock;
+use tokio::time::sleep;
```
Ponewor, line 13: The way imports are grouped makes absolutely no sense to me ;)

Ponewor (changes requested):

macher259:
Now all tests are passing.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        session.query(schema, ()).await?;
+        session.await_schema_agreement().await?;
+
+        Ok(())
+    }
+}
+
+fn get_default_generation_pk() -> Vec<u8> {
+    vec![0]
```
piodul, line 119: Why not just use an empty vec? IMO it differentiates itself better from other generations (empty vs non-empty) and should resolve that one issue with adding `42` to the stream number.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        let last_generation = cp_saver.load_last_generation().await.unwrap().unwrap();
+        assert_eq!(last_generation, good_generation);
+    }
+
+    // Creates NUMBER_OF_PRODUCERS producers that create unique checkpoints.
+    // Checks whether `load_last_generation` will really return the largest value of seen generations.
+    // Also checks whether number of checkpoints in the table is equal to the number of tasks making checkpoints + one special for generations.
+    #[tokio::test]
+    async fn test_multiple_producers() {
```
piodul, line 360: This test also seems to suffer from the timing issues. I'm not sure that we need such a complicated test; it seems to try and mimic how the library will save checkpoints, and IMO it will be better to write an integration test and use the actual CDCLogReader for that.

In this module, I'd rather see simpler and deterministic tests, each of them testing one characteristic of the checkpoint saver interface. For example:

1. Check that after calling `save_checkpoint` multiple times for one stream `load_last_checkpoint` will return the last checkpoint,
2. Check that after calling `save_checkpoint` once for each stream you will be able to retrieve the same checkpoints for each stream with `load_last_checkpoint`,
3. A similar test to 1) but operating on generations, not on checkpoints,
4. Etc.?

As for the integration test, we can use the end-to-end framework. The test can run a small number of rounds, in each, first write some data and then run the cdc log reader up to a timestamp after the current time. The test cdc log reader should use the checkpoint saver to restore from the previously saved progress, save progress periodically and also do it when the cdc log reader is stopped. At the end, we should make sure that all rows were read and each one of them was read only once (the framework and the test cdc log reader already does that).

piodul (changes requested):
Left some comments.

The integration of the checkpoint maker with CDCLogReader and StreamReader - looking forward to seeing it.

scylla-cdc/src/checkpoints.rs
macher259: I have changed return type to `StreamId` as I thought it would be more elegant than using bare vector and returning an empty stream id results in: `Database returned an error: The query is syntatically correct but invalid, Error message: Key may not be empty'`.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+        )
+    }
+
+    #[tokio::test]
+    async fn test_save_generation_multiple_times() {
+        const N: usize = 20;
+        let (session, table_name, cp_saver) = setup().await;
+
+        let mut generation = GenerationTimestamp {
+            timestamp: chrono::Duration::seconds(0),
```
Ponewor, line 290: `chrono::Duration::zero()`

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+        );
+    }
+
+    #[tokio::test]
+    async fn test_save_checkpoint_multiple_streams() {
+        const N_OF_IDS: u8 = 200;
+
+        let (_, _, cp_saver) = setup().await;
+
+        let mut checkpoints = Vec::<Checkpoint>::new();
```
Ponewor, line 319: You can allocate required memory using `with_capacity` and later store elements in corresponding fields instead of pushing.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+    checkpoint_table: String,
+    make_checkpoint_stmt: PreparedStatement,
+}
+
+impl CDCDefaultCheckpointSaver {
+    /// Creates new [`CDCDefaultCheckpointSaver`].
+    /// If given `None` value as table name,
+    /// it uses default value `"CDC_CHECKPOINT"`.
+    pub async fn new(
+        session: &Arc<Session>,
```
piodul, line 74: Style tip: if you have a function which takes ownership of a shared pointer - like here, you are cloning it and putting it into a struct - then it's better to take `Arc<T>` by value instead of a reference.

Moreover, sometime this lets you avoid a clone, e.g. the caller creates an `Arc<T>`, passes it by value to the function and no clones are made.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+pub struct CDCDefaultCheckpointSaver {
+    session: Arc<Session>,
+    checkpoint_table: String,
+    make_checkpoint_stmt: PreparedStatement,
+}
+
+impl CDCDefaultCheckpointSaver {
+    /// Creates new [`CDCDefaultCheckpointSaver`].
+    /// If given `None` value as table name,
+    /// it uses default value `"CDC_CHECKPOINT"`.
```
piodul, line 72: I'd rather require the user to be explicit about the table name.

I also think it is a bit uncommon to use `Option` for optional arguments with defaults... It's more common to have different constructors with more or less arguments (e.g. `new` vs. `new_with_checkpoint_saver`) or a builder pattern.

scylla-cdc/src/checkpoints.rs
piodul: I meant it to be a non-static method, so that it takes `self` and you don't have to explicitly pass `session` and `checkpoint_table`.

scylla-cdc/src/checkpoints.rs
piodul: The docstring should also mention that `new` creates the table if it didn't exist before.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+    async fn load_last_checkpoint(
+        &self,
+        stream_id: &StreamID,
+    ) -> anyhow::Result<Option<chrono::Duration>>;
+}
+
+/// Default implementation for [`CDCCheckpointSaver`] trait.
+pub struct CDCDefaultCheckpointSaver {
+    session: Arc<Session>,
+    checkpoint_table: String,
```
piodul, line 65: When the library switches to the new generation the current set of streams changes, so the new stream readers will save progress in a new set of partitions while the partitions for previous generation are left behind - so the table will grow every time there is a generation change.

We need to add another parameter: TTL. You can write data to scylla with `USING TTL = X` where `X` is the number of seconds after the row will "disappear" and scylla will eventually be able to reclaim the space it uses.

As generation changes aren't supposed to be frequent, you can provide a default TTL of 7 days.

scylla-cdc/src/log_reader.rs
```diff
@@ -350,18 +409,45 @@ impl CDCLogReaderBuilder {
...
+        }
+        let mut start_timestamp = self.start_timestamp;
+        if self.should_load_progress {
+            if let Some(generation) = checkpoint_saver
+                .as_ref()
+                .unwrap()
+                .load_last_generation()
+                .await?
+            {
+                start_timestamp = generation.timestamp;
```
piodul, line 431: I think there is a potential problem. Consider the following situation:

1. You created a cluster a week ago and didn't make any topology changes since,
2. You start the application with progress saving turned on and tell it to start reading from the `now` timestamp,
3. The application saves the generation but crashes immediately after that, without saving information about progress of any of the streams,
4. You restart the application, it loads the generation timestamp (which is set to a week ago), the stream readers see that no progress was saved for them and they start reading from the beginning of the generation.

I think we should store both the start timestamp and the generation timestamp in the progress table.

piodul (changes requested):

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,352 @@
...
+                stream_id: StreamID { id: vec![0, i] },
+                generation: GenerationTimestamp {
+                    timestamp: chrono::Duration::max_value(),
+                },
+            };
+
+            checkpoints.push(checkpoint);
+        }
+
+        for i in 0..N_OF_IDS {
```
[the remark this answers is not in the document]
Ponewor: I don't get why do we even need that loop at all. Can't we call `cp_saver.save_checkpoint(...)` in the previous loop?

scylla-cdc/src/log_reader.rs
macher259: Okay, I understand why `start_timestamp = generation.timestamp;` is a bad idea, but I think setting `start_timestamp` to `max(last_generation, start_timestamp` should be enough instead of storing the start timestamp. Am I missing something here?

piodul: OK, we can use `max(last_generation, start_timestamp)`.

scylla-cdc/src/stream_reader.rs
```diff
@@ -63,7 +78,34 @@ impl StreamReader {
         let mut window_begin = self.lower_timestamp;
         let window_size = chrono::Duration::from_std(self.window_size)?;
         let safety_interval = chrono::Duration::from_std(self.safety_interval)?;
+        let mut checkpoint = Checkpoint {
+            timestamp: window_begin.to_std()?,
+            stream_id: self.stream_id_vec[0].clone(),
```
piodul, line 73: If the stream reader is changed to fetch progress from multiple streams, then you will store progress of all those streams in a single checkpoint named by the first stream. While this will work, I dont like it because if we change how the streams are grouped then the application won't correctly restore progress after upgrading the library.

I think we should have a separate checkpoint for each stream. Maybe the task in `start_saving_checkpoint` could keep a list of stream IDs and update several checkpoints instead of only one? As far as restoring progress is concerned, we can start with an easy route: read checkpoints of all streams from the group and start reading all streams from the earliest checkpoint.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,366 @@
...
+    pub stream_id: StreamID,
+    pub generation: GenerationTimestamp,
+}
+
+/// Spawns a new task that periodically saves checkpoints.
+/// Takes vector of tracked by a [`StreamReader`] stream IDs.
+/// Receives value of the last checkpoint via channel.
+/// We are using [`tokio::sync::watch`] channel, because we only need the latest value.
+/// Returns handle to check on this newly created task.
+pub fn start_saving_checkpoints(
```
piodul, line 26: Shouldn't be `pub`, at most `pub(crate)`. This is not a function for the users.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,366 @@
...
+    receiver: tokio::sync::watch::Receiver<Checkpoint>,
+    saving_period: Duration,
+) -> RemoteHandle<anyhow::Result<()>> {
+    let (fut, handle) = async move {
+        loop {
+            for stream in tracked_streams.iter() {
+                let mut checkpoint = receiver.borrow().clone();
+                checkpoint.stream_id = stream.clone();
+
+                checkpoint_saver.save_checkpoint(&checkpoint).await?;
```
piodul, line 38: If `save_checkpoint` returns an error, the tokio task will return an error and progress will no longer be saved for this set of streams. Moreover, the error is returned in the handle, but the handle is discarded and error is not checked.

Please change this function so that it doesn't stop after encountering an error - instead, log it with `warn!` level (use the `tracing` crate which was introduced by another PR).

piodul (changes requested):
Sorry for the delay. One more thing needs to be fixed, but then I think we can merge it.

scylla-cdc/src/checkpoints.rs
piodul: You can leave it as a docstring.

scylla-cdc/src/checkpoints.rs
```diff
@@ -0,0 +1,372 @@
...
+/// Takes vector of tracked by a [`StreamReader`] stream IDs.
+/// Receives value of the last checkpoint via channel.
+/// We are using [`tokio::sync::watch`] channel, because we only need the latest value.
+/// Returns handle to check on this newly created task.
+pub(crate) fn start_saving_checkpoints(
+    tracked_streams: Vec<StreamID>,
+    checkpoint_saver: Arc<dyn CDCCheckpointSaver>,
+    receiver: tokio::sync::watch::Receiver<Checkpoint>,
+    saving_period: Duration,
+) -> RemoteHandle<anyhow::Result<()>> {
```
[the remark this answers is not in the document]
piodul: ...but that's just a nitpick, we can leave it for a followup.

piodul (approved):

PR #71, opened 2022-04-21
Make replicator executable

Ponewor:
As in the title - this PR attempts to make replicator executable.

We follow the interface from Java and Go versions of the library (at least when it comes to the options we support at the moment).

A few questions:

1. What are the values for `window_size`, `safety_interval` and `sleep_interval` we should use here?
2. Am I assuming correctly that the replicator starts to replicate from the beginning - timestamp 0? This will probably be changed anyway once @macher259 completes progress saving task, right?
3. The `end_timestamp` argument in CDCLogReader looks definitely awkard. IIRC @Gor027 you were working on removing it, right? How is it going?
4. Should we handle SIGINTs gracefully? Is there something to be done here or the SIGINT drops the handle by itself?

### DIFF
--- a/scylla-cdc-replicator/Cargo.toml
+++ b/scylla-cdc-replicator/Cargo.toml
@@ -15,3 +15,4 @@ anyhow = "1.0.48"
 itertools = "0.10.3"
 uuid = {version = "1.0.0", features = ["v1"]}
 futures-util = "0.3.19"
+tracing = "0.1.34"
--- a/scylla-cdc-replicator/Cargo.toml
+++ b/scylla-cdc-replicator/Cargo.toml
@@ -16,3 +16,5 @@ itertools = "0.10.3"
 uuid = {version = "1.0.0", features = ["v1"]}
 futures-util = "0.3.19"
 tracing = "0.1.34"
+chrono = "0.4.19"
+clap = { version = "3.1.10", features = ["derive"] }
--- a/scylla-cdc-replicator/src/main.rs
+++ b/scylla-cdc-replicator/src/main.rs
@@ -1,29 +1,94 @@
 mod replication_tests;
+mod replicator;
 pub mod replicator_consumer;
 
-use replicator_consumer::ReplicatorConsumerFactory;
-use scylla::SessionBuilder;
-use std::error::Error;
-use std::sync::Arc;
 use std::time::Duration;
 
+use clap::Parser;
+use futures_util::StreamExt;
+use tokio::select;
+
+use crate::replicator::Replicator;
+
+#[derive(Parser)]
+struct Args {
+    /// Keyspace name
+    #[clap(short, long)]
+    keyspace: String,
+
+    /// Table names provided as a comma delimited string
+    #[clap(short, long)]
+    table: String,
+
+    /// Address of a node in source cluster
+    #[clap(short, long)]
+    source: String,
+
+    /// Address of a node in destination cluster
+    #[clap(short, long)]
+    destination: String,
+
+    /// Window size in seconds
+    #[clap(long, default_value_t = 60.)]
+    window_size: f64,
+
+    /// Safety interval in seconds
+    #[clap(long, default_value_t = 30.)]
+    safety_interval: f64,
+
+    /// Sleep interval in seconds
+    #[clap(long, default_value_t = 10.)]
+    sleep_interval: f64,
+
+    /// Start datetime as RFC 3339 formatted string
+    #[clap(long)]
+    start_datetime: Option<String>,
+}
+
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn Error>> {
-    println!("Hello");
-    tokio::time::sleep(Duration::from_secs(1)).await;
+async fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let tables: Vec<_> = args.table.split(',').map(|s| s.to_string()).collect();
+    assert!(!tables.is_empty(), "no tables were provided");
+    let start_timestamp = chrono::Duration::milliseconds(match args.start_datetime {
+        Some(s) => chrono::DateTime::parse_from_rfc3339(&s)?.timestamp_millis(),
+        None => chrono::Local::now().timestamp_millis(),
+    });
+    let sleep_interval = Duration::from_secs_f64(args.sleep_interval);
+    let mut result = Ok(());
 
-    // Create a factory to remove the warning about unused new() function.
-    let _factory = ReplicatorConsumerFactory::new(
-        Arc::new(
-            SessionBuilder::new()
-                .known_node("127.0.0.1:9042")
-                .build()
-                .await?,
-        ),
-        "ks".to_string(),
-        "t".to_string(),
-    );
+    let (mut replicator, mut handles) = Replicator::new(
+        args.source,
+        args.keyspace.clone(),
+        tables.clone(),
+        args.destination,
+        args.keyspace,
+        tables,
+        start_timestamp,
+        Duration::from_secs_f64(args.window_size),
+        Duration::from_secs_f64(args.safety_interval),
+        sleep_interval,
+    )
+    .await?;
 
-    println!("Replicator");
-    Ok(())
+    select! {
+        _ = tokio::signal::ctrl_c() => {
+            println!("Shutting down...");
+        }
+        res = handles.next() => {
+            // The only way this could have happened is an error in on of the CDCLogReaders so we
+            // stop all of them.
+            result = res.unwrap();
+            println!("{:?}", result);
+        }
+    }
+
+    replicator.stop();
+    while let Some(res) = handles.next().await {
+        if res.is_err() {
+            result = res;
+            println!("{:?}", result);
+        }
+    }
+    result
 }
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -11,6 +11,7 @@ use scylla::prepared_statement::PreparedStatement;
 use scylla::query::Query;
 use scylla::transport::topology::{CollectionType, ColumnKind, CqlType, Table};
 use scylla::Session;
+use tracing::warn;
 
 use scylla_cdc::consumer::*;
 
@@ -836,7 +837,7 @@ impl Consumer for ReplicatorConsumer {
             OperationType::RowRangeDelInclLeft => self.delete_row_range_left(data, true),
             OperationType::RowRangeDelExclRight => self.delete_row_range_right(data, false).await?,
             OperationType::RowRangeDelInclRight => self.delete_row_range_right(data, true).await?,
-            _ => todo!("This type of operation is not supported yet."),
+            op => warn!("This type of operation - {:?} - is not supported yet.", op),
         }
 
         Ok(())
--- a/scylla-cdc-replicator/src/replicator.rs
+++ b/scylla-cdc-replicator/src/replicator.rs
@@ -1,0 +1,62 @@
+use std::sync::Arc;
+use std::time;
+
+use futures_util::future::RemoteHandle;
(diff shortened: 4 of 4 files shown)


piodul:
> 1. What are the values for `window_size`, `safety_interval` and `sleep_interval` we should use here?

Let's use the defaults from scylla-cdc-go. I believe we have different nomenclature - perhaps we should change it for consistency - but for now, it should be:

- `window_size`, equivalent `QueryTimeWindowSize` (60s)
- `safety_interval`: equivalent `ConfidenceWindowSize` (30s)
- `sleep_interval`: no direct equivalent because scylla-cdc-go uses different delays in case a poll fails, but let's use 10s by default.

It would be nice to be able to change those through command line arguments.

> 2. Am I assuming correctly that the replicator starts to replicate from the beginning - timestamp 0? This will probably be changed anyway once @macher259 completes progress saving task, right?

Yes to both. It should start from the beginning and should resume if the progress was saved.

> 3. The `end_timestamp` argument in CDCLogReader looks definitely awkard. IIRC @Gor027 you were working on removing it, right? How is it going?
> 4. Should we handle SIGINTs gracefully? Is there something to be done here or the SIGINT drops the handle by itself?

AFAIK if you don't register a signal handler, the process is aborted on SIGINT. You can use this: https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html to wait for a ctrlc and handle it gracefully (e.g. let the replicator read up to the current timestamp). After receiving the second ctrlc signal, you can proceed with terminating the process forcefully.

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,58 @@
...
+    destination: String,
+
+    #[clap(long, default_value_t = 60)]
+    window_size: u16,
+
+    #[clap(long, default_value_t = 30)]
+    safety_interval: u16,
+
+    #[clap(long, default_value_t = 10)]
+    sleep_interval: u16,
```
piodul, line 31: Please make it more clear what are the units here. Moreover, if it's seconds then I think a `double` would be a better choice so that sub-second intervals can be specified.

Maybe `clap` allows to just parse `Duration`s?

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,58 @@
...
+    keyspace: String,
+
+    #[clap(short, long)]
+    table: String,
+
+    #[clap(short, long)]
+    source: String,
+
+    #[clap(short, long)]
+    destination: String,
```
piodul, line 22: A short description of the flags that shows on `--help` in the CLI would be appreciated.

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,58 @@
...
+        Duration::from_secs(args.safety_interval as u64),
+        Duration::from_secs(args.sleep_interval as u64),
+    )
+    .await?;
+
+    tokio::signal::ctrl_c().await.unwrap();
+
+    println!("Shutting down...");
+
+    replicator.stop();
```
piodul, line 55: `stop()` does not wait for the replicator to finish, it just sets the upper timestamp so that the log reader stops after reaching it. The whole program will stop immediately if the `main` function returns, regardless of other running tasks.

If you want to gracefully wait until the replicator finishes, you should `.await` then handle returned from the replicator.

piodul (changes requested):

scylla-cdc-replicator/src/main.rs
Ponewor: Not `Duration`s per se but Duration wrapped in custom type with some traits implemented.
This of course requires parsing command line inputs into durations. Do you have any specific format in mind? Just a finite decimal representation of a number of seconds?

piodul: Hmm, no, I didn't have any specific format in mind. I thought that maybe Rust's standard library makes it possible to parse durations in some format, but if not then let's stick to `f64` and interpret as seconds. No need to write any custom wrappers.

scylla-cdc/src/log_reader.rs
```diff
@@ -184,10 +185,10 @@ impl CDCReaderWorker {
...
                         .iter()
-                        .map(|reader| {
+                        .cartesian_product(self.table_names.clone())
+                        .map(|(reader, table_name)| {
                             let reader = Arc::clone(reader);
                             let keyspace = self.keyspace.clone();
-                            let table_name = self.table_name.clone();
                             let factory = Arc::clone(&self.consumer_factory);
                             tokio::spawn(async move {
                                 let consumer = factory.new_consumer().await;
```
piodul, line 194: The consumer should be made aware from which table it consumes data. The new_consumer() method should accept the table/keyspace name.

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,65 @@
...
+    let args = Args::parse();
+    let source_tables = args.table.split(',').map(|s| s.to_string()).collect();
+
+    let (mut replicator, handle) = Replicator::new(
+        args.source,
+        args.keyspace.clone(),
+        source_tables,
+        args.destination,
+        args.keyspace,
+        args.table,
```
piodul, line 52: This doesn't look correct. You adjusted the CDC log reader to be able to read from multiple tables at the same time, but the replicator consumer still supports only one output table. Besides, you are giving it a comma-separated string of table names as the output table, which is guaranteed not to work when more than source table is specified.

Please adjust the replicator so that it supports multiple tables + add a test for it. If you wish, we can postpone it to a followup (but, in such case, please remove the ability to specify more than one source table for the replicator).

piodul (changes requested):

scylla-cdc/src/log_reader.rs
Ponewor: Are you sure you meant the source table? It's not a bad idea but I don't see how it's relevant here.
Maybe you meant the destination table? That would definitely be useful. But when I look at the traits we've defined
```rust
#[async_trait]
pub trait Consumer: Send {
    async fn consume_cdc(&mut self, data: CDCRow<'_>) -> anyhow::Result<()>;
}

#[async_trait]
pub trait ConsumerFactory: Sync + Send {
    async fn new_consumer(&self) -> Box<dyn Consumer>;
}
```

I don't see how a method like
```rust
async fn new_consumer(&self, dest_keyspace: &str, dest_table: &str) -> Box<dyn Consumer>
```
would fit such a general interface. For example `PrinterConsumer` doesn't have a need for destination table/keyspace.

Maybe we could have multiple `ReplicatorConsumerFactories`?

Ponewor: I've just pushed the implementation with multiple factories for inspection.

piodul: Sorry for the late response.

The more I think about the multi-table support the less sense I think it makes. I'm not sure there will be much benefit at the moment in supporting it natively in the CDCLogReader vs. just running multiple instances of the CDCLogReader (I just noticed that the scylla-cdc-java replicator does that).

Although running multiple CDCLogReaders duplicates work related to generation fetching, it is a tiny amount of work and complicating the CDCLogReader in order to save work isn't worth it IMO.

Let's drop the multi-table support and just change the replicator to run multiple instances of the CDCLogReader.

Ponewor: A question then:

How many tables do we expect? Can the handles from `CDCLogReader`s be awaited with something like `join_all` or do we need use `FuturesUnordered` again?

piodul: I don't think the number of tables will be big, but it is a user-controlled parameter so `FuturesUnordered` won't hurt here.

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,68 @@
...
+        Duration::from_secs_f64(args.sleep_interval),
+    )
+    .await?;
+
+    tokio::signal::ctrl_c().await.unwrap();
+
+    println!("Shutting down...");
+
+    replicator.stop();
+    while handles.next().await.is_some() {}
```
piodul, line 66: I think there is a problem here... If one of the replicators stops due to an error, the whole application won't notice unless the user presses Ctrl+C. The application should react immediately if this happens, i.e. stop other replicators and - ideally - wait until remaining replicators complete.

Moreover, you are not reporting the errors in any way, i.e. no error is returned from main and nothing is printed. Unless there is some logging in the replicator (which I'm not aware of) you won't even know why it stopped and if it was due to a failure or not.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -11,7 +11,6 @@ use scylla::prepared_statement::PreparedStatement;
 use scylla::query::Query;
 use scylla::transport::topology::{CollectionType, ColumnKind, CqlType, Table};
 use scylla::Session;
-
```
piodul, line 14: Unnecessary whitespace change?

piodul (changes requested):

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,93 @@
...
+    )
+    .await?;
+
+    loop {
+        select! {
+            _ = tokio::signal::ctrl_c() => {
+                println!("Shutting down...");
+                replicator.stop();
+                while handles.next().await.is_some() {}
+                return Ok(())
```
piodul, line 80: Nit: missing semicolon at the end of line

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,93 @@
...
+            res = handles.next(), if !handles.is_terminated() => {
+                let res = res.unwrap();
+                if res.is_err() {
+                    replicator.stop();
+                    while handles.next().await.is_some() {}
+                    return res;
+                }
+            }
+        }
+        sleep(sleep_interval).await;
```
piodul, line 91: Why is this sleep needed? It looks like we can arrive here only when `handles.next()` returns a non-error, which happens when one of the CDC log readers finishes reading. Won't it unnecessarily delay shutdown?

scylla-cdc-replicator/Cargo.toml
```diff
@@ -14,4 +14,5 @@ async-trait = "0.1.51"
 anyhow = "1.0.48"
 itertools = "0.10.3"
 uuid = {version = "1.0.0", features = ["v1"]}
-futures-util = "0.3.19"
\ No newline at end of file
+futures-util = "0.3.19"
+log = "0.4.17"
```
piodul, line 18: I'd recommend using the [`tracing`](https://docs.rs/tracing/latest/tracing/) crate instead. It integrates with asynchronous code better, and the rust driver uses it, too.

scylla-cdc-replicator/src/main.rs
piodul: Now that I think about it... Is this code even reachable? The upper timestamp is not set for the log readers, so I don't think they can decide to stop by themselves.

Ponewor: You're right

Ponewor: In fact the loop itself is useless.

scylla-cdc-replicator/src/main.rs
```diff
@@ -1,29 +1,86 @@
...
+
+    select! {
+        _ = tokio::signal::ctrl_c() => {
+            println!("Shutting down...");
+            replicator.stop();
+            while handles.next().await.is_some() {}
+            Ok(())
+        }
+        res = handles.next() => {
+            replicator.stop();
```
piodul, line 81: I think we shouldn't stop here unconditionally - we should tell other consumers to stop only if we got an error.

Actually, now that I have wrote the first sentence, I realized that the replicator is supposed to stop only in case of an error. It's not immediately obvious IMO so please add a comment about it.

piodul (approved):

PR #76, opened 2022-04-27
Fix e2e_tests error resulting in not comparing expected and actual operations lengths after a difference was found in previous loop iterations

Ponewor:
As in the title.
[https://github.com/piodul/scylla-cdc-rust/pull/46/files#r859401423](url).

Also: refactor the tests to use uniquely named keyspaces and other tools from tests_utilites while we're at it.

### DIFF
--- a/scylla-cdc/src/e2e_tests.rs
+++ b/scylla-cdc/src/e2e_tests.rs
@@ -11,6 +11,7 @@ mod tests {
     use scylla::prepared_statement::PreparedStatement;
     use scylla::{Session, SessionBuilder};
     use std::collections::{HashMap, VecDeque};
+    use std::convert::identity;
     use std::hash::Hash;
     use std::sync::Arc;
     use std::time;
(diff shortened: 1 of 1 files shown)


piodul (changes requested):

Ponewor:
@piodul

scylla-cdc/src/e2e_tests.rs
```diff
@@ -290,48 +290,38 @@ mod tests {
...
-                if is_ok && expected_operations.len() != actual_operations.len() {
-                    eprintln!(
-                        "Too many read operations for primary key {:?}. Operations left: {}",
-                        pk,
-                        actual_operations.len()
-                    );
-                    is_ok = false;
-                }
+            for pk in self.performed_operations.keys() {
+                eprintln!("Unexpected primary key {:?}", pk);
```
piodul, line 289: The message is wrong - we did expect `pk` to be seen by the reader but it wasn't (IIUC) - also, it is exactly the same as the message on line `299/297` which will only make debugging harder when it pops up because it won't be possible to tell which line actually printed it. Please improve it.

scylla-cdc/src/e2e_tests.rs
```diff
@@ -1,22 +1,25 @@
...
-    use scylla::frame::response::result::CqlValue;
-    use scylla::frame::value::{Value, ValueTooBig};
-    use scylla::prepared_statement::PreparedStatement;
-    use scylla::{Session, SessionBuilder};
     use std::collections::{HashMap, VecDeque};
     use std::convert::identity;
     use std::hash::Hash;
     use std::sync::Arc;
     use std::time;
+
```
piodul, line 8: The second commit should be sent in a separate PR. It's a refactor unrelated to the original problem this PR is supposed to solve.

piodul (commented):
The first commit looks good. Please resend the "Refactor: integrate e2e_tests with test_utilities" as a separate PR and remove it from the current one.

piodul (approved):

PR #77, opened 2022-04-27
Cdc log reader builder

### LINKED ISSUE #58: Log reader builder
The CDCLogReader (which will be created from CDCLogPrinter) has a constructor with a large number of parameters. Functions with a large number of arguments are in bad style, are hard to read and aren't forward compatible in case we decide to add more parameters.

Instead, we should provide a CDCLogReaderBuilder struct which starts with sensible defaults and has a separate method for setting each option. It should have a `.build()` method which consumers the builder, performs some validation of the operations and creates an instance of CDCLogReader.

The [SessionBuilder](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/transport/session_builder.rs) from the scylla-rust-driver can serve as an inspiration.

Gor027:
Resolves #58

This PR replaces CDC log reader and CDC reader worker constructors with too many arguments by adding CDC log reader builder.

### DIFF
--- a/scylla-cdc/src/log_reader.rs
+++ b/scylla-cdc/src/log_reader.rs
@@ -19,6 +19,7 @@ const DEFAULT_SLEEP_INTERVAL: i64 = SECOND_IN_MILLIS * 10;
 const DEFAULT_WINDOW_SIZE: i64 = SECOND_IN_MILLIS * 60;
 const DEFAULT_SAFETY_INTERVAL: i64 = SECOND_IN_MILLIS * 3;
 
+/// To create a new CDCLogReader instance please see documentation for [`CDCLogReaderBuilder`]
 pub struct CDCLogReader {
     // Tells the worker to stop
     // Usage of the "watch" channel will make it possible to change the the timestamp later,
@@ -163,6 +164,47 @@ impl CDCReaderWorker {
     }
 }
 
+/// CDCLogReaderBuilder is used to create new [`CDCLogReader`] instances.
+/// # Example
+///
+/// ```
+/// # use scylla_cdc::log_reader::{CDCLogReader, CDCLogReaderBuilder};
+/// # use scylla_cdc::consumer::{ConsumerFactory, Consumer, CDCRow};
+/// # use futures::future::RemoteHandle;
+/// # use scylla::{Session, SessionBuilder};
+/// # use std::sync::Arc;
+/// # use async_trait::async_trait;
+/// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+/// struct DummyConsumer;
+///
+/// #[async_trait]
+/// impl Consumer for DummyConsumer {
+///     async fn consume_cdc(&mut self, data: CDCRow<'_>) -> anyhow::Result<()> {
+///         // ... consume received data ...
+///         Ok(())
+///     }
+/// }
+///
+/// struct DummyConsumerFactory;
+///
+/// #[async_trait]
+/// impl ConsumerFactory for DummyConsumerFactory {
+///     async fn new_consumer(&self) -> Box<dyn Consumer> {
+///         Box::new(DummyConsumer)
+///     }
+/// }
+///
+/// let session = SessionBuilder::new().known_node("127.0.0.1:9042").build().await.unwrap();
+/// let (cdc_log_printer, handle): (CDCLogReader, RemoteHandle<anyhow::Result<()>>) = CDCLogReaderBuilder::new()
+///     .session(Arc::new(session))
+///     .keyspace("log_reader_ks")
+///     .table_name("log_reader_table")
+///     .consumer_factory(Arc::new(DummyConsumerFactory))
+///     .build()
+///     .await?;
+/// # Ok(())
+/// # }
+/// ```
 pub struct CDCLogReaderBuilder {
     session: Option<Arc<Session>>,
     keyspace: Option<String>,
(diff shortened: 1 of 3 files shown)


scylla-cdc/src/log_reader_builder.rs
```diff
@@ -0,0 +1,310 @@
...
+
+enum NewCDCLogReaderError {
+    TableName,
+    KeyspaceName,
+    Session,
+    ConsumerFactory,
+}
+
+impl NewCDCLogReaderError {
+    fn as_str(&self) -> &'static str {
```
piodul, line 32: It's more idiomatic to implement the [`Display`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html) trait and add `#[derive(Debug)]` for errors rather than introducing custom methods like that.

scylla-cdc/src/log_reader.rs
```diff
@@ -17,7 +17,7 @@ pub struct CDCLogReader {
     // Tells the worker to stop
     // Usage of the "watch" channel will make it possible to change the the timestamp later,
     // for example if somebody loses patience and wants to stop now not later
-    end_timestamp: tokio::sync::watch::Sender<chrono::Duration>,
+    end_timestamp: tokio::sync::watch::Sender<Option<chrono::Duration>>,
```
piodul, line 5: Perhaps a better idea than changing the type to `Option<Duration>` is to provide `Duration::max_value()` or something like that instead of `None`.

scylla-cdc/src/log_reader_builder.rs
```diff
@@ -0,0 +1,310 @@
...
+impl Default for CDCLogReaderBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+struct CDCReaderWorker {
+    session: Option<Arc<Session>>,
+    keyspace: Option<String>,
+    table_name: Option<String>,
```
piodul, line 198: Why did you change the field types to `Option` in the reader worker? The reader builder is supposed to validate that they are not `None`, so no need to keep them as `Option`s in the reader worker. This only introduces completely unnecessary unwrapping - future readers of the code will wonder why it is needed and if it can actually fail or not.

scylla-cdc/src/log_reader_builder.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    pub fn consumer_factory(mut self, consumer_factory: Arc<dyn ConsumerFactory>) -> Self {
+        self.consumer_factory = Some(consumer_factory);
+        self
+    }
+
+    pub async fn build(self) -> anyhow::Result<(CDCLogReader, RemoteHandle<anyhow::Result<()>>)> {
+        if self.table_name.is_none() {
+            return Err(anyhow!(NewCDCLogReaderError::TableName.as_str()));
```
piodul, line 152: The new error type is completely unnecessary if you are always converting it to string anyway. Either return the new error type in the `Result<(..., ...), NewCDCLogReaderError>`, or get rid of it and inline the strings.

scylla-cdc/src/log_reader_builder.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        if self.keyspace.is_none() {
+            return Err(anyhow!(NewCDCLogReaderError::KeyspaceName.as_str()));
+        }
```
piodul, line 157: As I previously mentioned, you are validating that those `Option`s are not `None` but are putting them as options into the `CDCReaderWorker` anyway. I recommend de-optioning them and using the following pattern:

```rust
let keyspace = self.keyspace.ok_or(NewCDCLogReaderError::KeyspaceName)?;
/* or, if you are going to use anyhow */
let keyspace = self.keyspace.ok_or_else(|| anyhow!("failed to create the cdc reader: missing keyspace name"));

/* ... */

let cdc_reader_worker = CDCReaderWorker {
    /* ... */
    keyspace, // No need for "keyspace: keyspace,", just a single "keyspace" works as well
    /* ... */
};
```

piodul (changes requested):

scylla-cdc/src/log_reader.rs
```diff
@@ -13,6 +14,11 @@ use crate::consumer::ConsumerFactory;
 use crate::stream_generations::GenerationFetcher;
 use crate::stream_reader::StreamReader;
 
+const SECOND_IN_MILLIS: i64 = 1_000;
+const SLEEP_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
+const WINDOW_SIZE: i64 = SECOND_IN_MILLIS / 10 * 3;
+const SAFETY_INTERVAL: i64 = SECOND_IN_MILLIS / 10;
```
piodul, line 20: Let's prefix those constants with `DEFAULT_` to indicate that those are the default values for those parameters.

scylla-cdc/src/log_reader.rs
```diff
@@ -214,3 +162,147 @@ impl CDCReaderWorker {
...
+pub struct CDCLogReaderBuilder {
+    session: Option<Arc<Session>>,
+    keyspace: Option<String>,
+    table_name: Option<String>,
+    start_timestamp: chrono::Duration,
+    end_timestamp: chrono::Duration,
+    window_size: time::Duration,
+    safety_interval: time::Duration,
+    sleep_interval: time::Duration,
+    readers: Vec<Arc<StreamReader>>,
```
piodul, line 175: The readers field is unnecessary. The new CDCLogReaderBuider should start with no readers.

piodul (changes requested):

piodul (changes requested):

Gor027:
ping @piodul Is this PR ready to be merged?

piodul (approved):

scylla-cdc/src/log_reader.rs
```diff
@@ -22,38 +29,8 @@ pub struct CDCLogReader {
...
-            end_timestamp_receiver,
-            consumer_factory,
-        );
-        let (fut, handle) = async move { worker.run().await }.remote_handle();
-        tokio::task::spawn(fut);
-        let printer = CDCLogReader {
-            end_timestamp: end_timestamp_sender,
-        };
-        (printer, handle)
+    pub fn new(end_timestamp: tokio::sync::watch::Sender<chrono::Duration>) -> Self {
```
piodul, line 32: This method shouldn't be public anymore, it is useless for the users. Please send a followup which fixes the visibility (and the comment, as it is also incorrect).

Issue #73, opened 2022-04-26
Reduce copy-paste in `PrecomputedQueries`


piodul:
The `PrecomputedQueries` struct has several methods which look up a prepared query from a hashmap, and - if it does not exist - prepares it before executing. All of them use a similar pattern which could be abstracted out.

One way to do it would be to introduce a type representing a per-query-type cache. It would keep a hashmap of prepared queries and a function which generates query string based on the query type.

Something like this:

```rust
struct SimpleQueryCache {
	queries: HashMap<String, PreparedQuery>,
	query_string_maker: Box<Fn(&str) -> String>,
}

impl SimpleQueryCache {
	pub async fn run_query(&mut self, session: Arc<Session>) -> Result<()> {
		/* ... */
	}
}
```

Ref: https://github.com/piodul/scylla-cdc-rust/pull/53#discussion_r852816487

PR #78, opened 2022-04-28
Refactor: replicator_consumer minor style improvements

Ponewor:
Just some meaningless nitpicks.


Ponewor:
@piodul

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -149,34 +143,29 @@ impl PrecomputedQueries {
...
+                    .dest_session
+                    .prepare(format!(
+                        "UPDATE {}.{} USING TTL ? SET {} = ? WHERE {}",
+                        self.destination_table_params.dest_keyspace_name,
+                        self.destination_table_params.dest_table_name,
+                        column_name,
+                        self.destination_table_params.keys_cond
+                    ))
+                    .await?,
+            );
```
piodul, line 158: There are two problems here:

1. Once again I'm pointing out that Rust is not Haskell and it has strict evaluation, not lazy - now, the argument to `or_insert` is _always_ evaluated, and the query is _always_ prepared, no matter if it was in the hashmap or not. The lazy counterpart to `or_insert` is `or_insert_with`, but you can't use it here because it doesn't support async/fallible initialization. Alternatively, you could pattern match on `Entry::Vacant` and initialize only then prepare the query, but see point 2).
2. Assuming that the problem in 1) does not exist, you are optimizing the less frequent case while pessimizing the common one. Now, when the statement is not in the map, you only need one lookup (just `entry` instead of `get_mut` + `entry`), but when the statement _is_ in the map, you are performing an additional allocation with `column_name.to_string()`, which the previous version didn't have as, due to the magic of traits, `get_mut` can accept a `&str`.

To sum up, I think we should drop this and similar changes in the commit. A better plan for these functions IMO is to deduplicate the logic like described in https://github.com/piodul/scylla-cdc-rust/issues/73.

piodul (changes requested):
Just one comment.

Ponewor:
Ok, I've removed those changes and @procrastinationfighter will deal with #73 ,

piodul (approved):

PR #80, opened 2022-04-28
Adjust uuid version to the one used in scylla-rust-driver

Ponewor:
UUID version in driver was changed and it causes some problems here.


Ponewor:
@piodul could you have a quick look at this one? It causes issues in other PRs...

piodul (approved):

PR #81, opened 2022-04-28
Enforce log reader stop after failure of one of the stream readers

Ponewor:
If my understanding is correct new generation can prolong a lifetime of stream readers and continuous calls to `CDCLogReader.stop_at` could even prevent them from stopping.


Ponewor:
@piodul

piodul (approved):

PR #82, opened 2022-04-29
Replicator: Cache queries through common interface

### LINKED ISSUE #73: Reduce copy-paste in `PrecomputedQueries`

procrastinationfighter:
Fixes #73

### DIFF
--- a/scylla-cdc-replicator/src/replicator_consumer.rs
+++ b/scylla-cdc-replicator/src/replicator_consumer.rs
@@ -14,6 +14,39 @@ use scylla::Session;
 
 use scylla_cdc::consumer::*;
 
+struct PreparedStatementCache {
+    queries: HashMap<String, PreparedStatement>,
+    query_string_maker: Box<dyn (Fn(&str) -> String) + Sync + Send>,
+}
+
+impl PreparedStatementCache {
+    pub fn new(query_string_maker: impl (Fn(&str) -> String) + Sync + Send + 'static) -> Self {
+        PreparedStatementCache {
+            queries: HashMap::new(),
+            query_string_maker: Box::new(query_string_maker),
+        }
+    }
+
+    pub async fn run_query(
+        &mut self,
+        session: &Arc<Session>,
+        values: &[impl Value],
+        timestamp: i64,
+        str_param: &str,
+    ) -> anyhow::Result<()> {
+        let query = match self.queries.get_mut(str_param) {
+            Some(query) => query,
+            None => self.queries.entry(str_param.to_string()).or_insert(
+                session
+                    .prepare((self.query_string_maker)(str_param))
+                    .await?,
+            ),
+        };
+
+        run_prepared_statement(session, query.clone(), values, timestamp).await
+    }
+}
+
 struct DestinationTableParams {
     session: Arc<Session>,
     keyspace_table_name: String,
@@ -27,10 +60,10 @@ struct PrecomputedQueries {
     insert_query: PreparedStatement,
     partition_delete_query: PreparedStatement,
     delete_query: PreparedStatement,
-    delete_value_queries: HashMap<String, PreparedStatement>,
-    overwrite_queries: HashMap<String, PreparedStatement>,
-    update_list_elements_queries: HashMap<String, PreparedStatement>,
-    update_map_or_set_elements_queries: HashMap<String, PreparedStatement>,
+    delete_value_queries: PreparedStatementCache,
+    overwrite_queries: PreparedStatementCache,
+    update_list_elements_queries: PreparedStatementCache,
+    update_map_or_set_elements_queries: PreparedStatementCache,
 }
 
 impl PrecomputedQueries {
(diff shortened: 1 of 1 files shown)


scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -79,17 +104,59 @@ impl PrecomputedQueries {
...
+        let update_map_or_set_elements_queries = PreparedStatementCache {
+            queries: HashMap::new(),
+            query_string_maker: Box::new(move |column_name| format!(
+                "UPDATE {tbl} USING TTL ? SET {cname} = {cname} + ?, {cname} = {cname} - ? WHERE {cond}",
+                tbl = c_ks_t,
+                cname = column_name,
+                cond = c_k_c,
+            )),
+        };
+
```
procrastinationfighter, line 153: Apparently, Rust does not have a convienient method for cloning items that we want to move into the closures: https://github.com/rust-lang/rfcs/issues/2407

procrastinationfighter:
@piodul

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -14,6 +14,32 @@ use scylla::Session;
 
 use scylla_cdc::consumer::*;
 
+struct PreparedStatementCache {
+    queries: HashMap<String, PreparedStatement>,
+    query_string_maker: Box<dyn (Fn(&str) -> String) + Sync + Send>,
+}
```
piodul, line 20: I suggest adding a `PreparedStatementCache::new` method so that you don't have to write `queries: HashMap::new()` every time.

Moreover, you can probably also avoid writing Box multiple times by using `impl` and boxing inside the constructor:

```rust
impl PreparedStatementCache {
    pub fn new(query_string_maker: impl (Fn(&str) -> String) + Sync + Send) -> Self {
        /* function contents left as an exercise for the reader */
    }
}
```

scylla-cdc-replicator/src/replicator_consumer.rs
piodul: Yes, I'm afraid there is no better way to do this than to clone needed items before creating the closure.

scylla-cdc-replicator/src/replicator_consumer.rs
```diff
@@ -79,17 +105,58 @@ impl PrecomputedQueries {
...
+        let update_map_or_set_elements_queries = PreparedStatementCache {
+            queries: HashMap::new(),
+            query_string_maker: Box::new(move |column_name| {
+                format!(
+                "UPDATE {tbl} USING TTL ? SET {cname} = {cname} + ?, {cname} = {cname} - ? WHERE {cond}",
+                tbl = c_ks_t,
+                cname = column_name,
+                cond = c_k_c,
+            )
+            }),
```
Ponewor, line 151: rustfmt is drunk or out of his mind

scylla-cdc-replicator/src/replicator_consumer.rs
procrastinationfighter: This definitions gives me  the following error:
```
error[E0310]: the parameter type `impl Fn(&str) -> String + Sync + Send` may not live long enough
  --> scylla-cdc-replicator/src/replicator_consumer.rs:26:33
   |
23 |     pub fn new(query_string_maker: impl (Fn(&str) -> String) + Sync + Send) -> Self {
   |                                    --------------------------------------- help: consider adding an explicit lifetime bound...: `impl Fn(&str) -> String + Sync + Send + 'static`
...
26 |             query_string_maker: Box::new(query_string_maker),
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...so that the type `impl Fn(&str) -> String + Sync + Send` will meet its required lifetime bounds
```
Is it ok to add `'static` lifetime? I'm not sure if it (changing the argument's lifetime) changes anything, since it's only a constraint for the argument.

piodul: Yes, requiring static is fine here. `'static` just means that the lifetime of the type is not constrained by other lifetimes. For example, it won't accept closures which capture variables from the stack by reference because those variables will be destroyed when the function goes out of scope (so they have non-static lifetime). In your case you are capturing owned Strings by `move` and the closure manages the lifetime of the strings itself, so `'static` will work perfectly fine.

piodul (approved):

PR #86, opened 2022-05-04
Make printer executable

### LINKED ISSUE #83: Improve the printer
As for now, the printer exists in a form of library. There are two things that should be done:

- Add prettier printing, because now it only prints the cql information from rows.
- Make it an executable, so that it can serve as an example of how to build an application using our library.

Ponewor:
Partially resolves #83


piodul (changes requested):
Looks good, but needs rebase and update. Although the PR has no conflicts, it won't build because the CDCLogReader is now created through CDCLogReaderBuilder.

piodul (approved):

PR #90, opened 2022-05-10
Create a tutorial for the library

### LINKED ISSUE #75: Library documentation
The library needs comprehensive documentation.

- Everything in the public interface (i.e. types, functions, traits) should contain documentation docstrings.
- We need a tutorial/guide introducing new users to the library and its features. It can be just a single .md file. The main page of the crate's documentation needs to contain a link to this tutorial/guide.

procrastinationfighter:
Fixes #75 
This PR adds a file tutorial.md which contains a short tutorial about how to use our library. It also fills missing documentation, adds the project's license and a readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,46 @@
+# scylla-cdc-rust
+Scylla-cdc-rust is a library that enables consuming the 
+[Scylla Change Data Capture Log](https://docs.scylladb.com/using-scylla/cdc/)
+in Rust applications. 
+The library automatically and transparently handles errors and topology changes 
+of the underlying Scylla cluster. 
+Thanks to that, the API allows the user to read the CDC log without having to deeply understand 
+the internal structure of CDC.
+
+It is recommended to get familiar with the documentation of CDC first, 
+in order to understand the concept: 
+<https://docs.scylladb.com/using-scylla/cdc/>
+
+The library was written in pure Rust,
+using [Scylla Rust Driver](https://crates.io/crates/scylla)
+and [Tokio](https://crates.io/crates/tokio).
+
+## Getting started
+The best place to get started with the library is the [tutorial](tutorial.md). 
+You can also check the [Scylla University lessons](https://university.scylladb.com/courses/scylla-operations/lessons/change-data-capture-cdc/)
+to learn more about the CDC.
+
+## Examples
+The repository also contains two exemplary applications:
+
+- [Printer](scylla-cdc-printer): prints all changes from CDC log for a selected table.
+- [Replicator](scylla-cdc-replicator): replicates a table from one Scylla cluster to another by reading the CDC log.
+
+## Contact
+Use the GitHub Issues to report bugs or errors. 
+You can also join [ScyllaDB-Users Slack channel](http://slack.scylladb.com/) and discuss on `#cdc` channel.
+
+## Useful links
+
+- [Scylla Rust Driver on GitHub](https://github.com/scylladb/scylla-rust-driver)
+- [Scylla Docs - Change Data Capture (CDC)](https://docs.scylladb.com/using-scylla/cdc/)
+- [Scylla University - Change Data Capture (CDC)](https://university.scylladb.com/courses/scylla-operations/lessons/change-data-capture-cdc/)
+- [ScyllaDB YouTube - Change Data Capture in Scylla](https://www.youtube.com/watch?v=392Nbfrq7Dg)
+- [ScyllaDB Blog - Using Change Data Capture (CDC) in Scylla](https://www.scylladb.com/2020/07/23/using-change-data-capture-cdc-in-scylla/)
+- [scylla-cdc-java - A library for Java](https://github.com/scylladb/scylla-cdc-java)
+- [scylla-cdc-go - A library for Go](https://github.com/scylladb/scylla-cdc-go)
+
+## License
+The library is licensed under Apache License 2.0. 
+You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0 
+or in the LICENSE.txt file in the repository.
--- a/scylla-cdc/src/consumer.rs
+++ b/scylla-cdc/src/consumer.rs
@@ -49,6 +49,15 @@ impl fmt::Display for OperationType {
     }
 }
 
+const STREAM_ID_NAME: &str = "cdc$stream_id";
+const TIME_NAME: &str = "cdc$time";
+const BATCH_SEQ_NO_NAME: &str = "cdc$batch_seq_no";
+const END_OF_BATCH_NAME: &str = "cdc$end_of_batch";
+const OPERATION_NAME: &str = "cdc$operation";
+const TTL_NAME: &str = "cdc$ttl";
+const IS_DELETED_PREFIX: &str = "cdc$deleted_";
+const ARE_ELEMENTS_DELETED_PREFIX: &str = "cdc$deleted_elements_";
+
 pub struct CDCRowSchema {
     // The usize values are indices of given values in the Row.columns vector.
     pub(crate) stream_id: usize,
@@ -70,15 +79,6 @@ pub struct CDCRowSchema {
     pub(crate) deleted_el_mapping: HashMap<String, usize>,
 }
 
-const STREAM_ID_NAME: &str = "cdc$stream_id";
-const TIME_NAME: &str = "cdc$time";
-const BATCH_SEQ_NO_NAME: &str = "cdc$batch_seq_no";
-const END_OF_BATCH_NAME: &str = "cdc$end_of_batch";
-const OPERATION_NAME: &str = "cdc$operation";
-const TTL_NAME: &str = "cdc$ttl";
-const IS_DELETED_PREFIX: &str = "cdc$deleted_";
-const ARE_ELEMENTS_DELETED_PREFIX: &str = "cdc$deleted_elements_";
-
 impl CDCRowSchema {
     pub fn new(specs: &[ColumnSpec]) -> CDCRowSchema {
         let mut stream_id = 0;
@@ -96,7 +96,6 @@ impl CDCRowSchema {
         // Hashmaps will have indices of data in a new vector without the hardcoded values.
         for (i, spec) in specs.iter().enumerate() {
             match spec.name.as_str() {
-                // spec.name is public since 0.3.0 driver version, it didn't work on 0.2.1.
                 STREAM_ID_NAME => stream_id = i,
                 TIME_NAME => time = i,
                 BATCH_SEQ_NO_NAME => batch_seq_no = i,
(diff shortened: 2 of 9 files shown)


procrastinationfighter:
@piodul

README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Examples
+The repository also contains two exemplary applications:
+
+- [Printer](scylla-cdc-printer): prints all changes from CDC log for a selected table.
+- [Replicator](scylla-cdc-replicator): replicates a table from one Scylla cluster to another by reading the CDC log.
+
+## Contact
+TODO
+
```
procrastinationfighter, line 31: What should we put here? Scylla-cdc-java directs to the `#cdc` channel on the slack.

piodul: Yes, it is a good idea to direct to the same slack channel here.

tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+    async fn new_consumer(&self) -> Box<dyn Consumer> {
+        Box::new(TutorialConsumer)
+    }
+}
+```
+
+### Adding shared state to the consumer
+A struct implementing the `Consumer` trait should also implement `Send` trait 
+and a struct implementing the `ConsumerFactory` trait should implement `Send` and `Sync` traits.
+Luckily, Rust implements these traits by default if all member variables of a struct implement them.
```
piodul, line 60: For context, as to _why_ those traits are required, we should mention that the consumer logic is run in separate tokio tasks and the runtime may schedule them on separate threads.

tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+
+If the consumers need to share some state, like a reference to an object, 
+they can be wrapped in an [Arc](<https://doc.rust-lang.org/std/sync/struct.Arc.html>).
```
piodul, line 63: An example of this would be nice here in order to illustrate the idea better. For example, you can add a shared counter which indicates the number of rows read so far, and use that number when printing to indicate the "ID" of the row.

We should also include a standard warning here that mutable state should be avoided as it requires synchronization (atomics, mutexes, etc.) which in turn reduces the speedup offered by tokio running the consumer logic on multiple cores.

tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+_________
+_________
+Time UUID: cc07e47a-d07c-11ec-ebc7-e1924abf3e16, operation type: RowUpdate
+pk: 1, ck: 2
+v: Some(8), was deleted: false
+vs: None, was deleted: false, deleted elements: []
+_________
+```
+
+<b>Note:</b> It is not guaranteed that the `CDCRow` struct will be alive after the function is terminated,
```
piodul, line 196: I think it could be phrased better. It's not that "it's not guaranteed that the CDCRow will be alive" - the CDCRow's lifetime _does not allow it to exist_ after the function terminates.

tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+        println!("Hello, scylla-cdc!");
+        Ok(())
+    }
+}
+```
+
+We'll cover on how to use the `CDCRow` structure in [Using CDCRow](tutorial.md#Using CDCRow).
+
+The library is going to create many instances of `TutorialConsumer`, so we also need to define
+a `ConsumerFactory` for them:
```
piodul, line 45: IMO it would be good to mention that the library creates a separate consumer for each cdc stream.

piodul (changes requested):
A topic not really explored in our documentation but which I think is worth explaining is the algorithm the library uses to read the CDC log. So, we should explain things like:

- The fact that the library reads one generation at a time, and reads all streams from that generation concurrently,
- Lifetime of the consumers (when they are created and when they are destroyed),
- How the window sizes, delays, etc. relate to how a single stream is being read.

I'm not sure what would be the best place (or places?) to put it, so I leave the choice up to you.

procrastinationfighter:
Besides applying your suggestions, I also added some module-level docstrings for the modules `stream_reader`, `log_reader` and `consumer`.

tutorial.md
```diff
@@ -0,0 +1,231 @@
...
+        counter.add(1);
+        println!("Row no.{}", counter.get());
+        Ok(())
+    }
+}
+```
+(an external crate [atomic_counter](https://docs.rs/atomic-counter/latest/atomic_counter/) was used in this example)
+
+<b>Note:</b> in general, keeping a mutable state in the `Consumer` is not recommended, 
+since it requires synchronization (i.e. a mutex),
```
piodul, line 91: I'm guessing that you added a mutex to the `CountingConsumer` so that this statement makes more sense? Atomics are more costly than non-atomic integers, especially if they are accessed from multiple threads at once, so the advice applies to them, too.

tutorial.md
```diff
@@ -0,0 +1,231 @@
...
+vs: None, was deleted: false, deleted elements: []
+_________
+```
+
+<b>Note:</b> CDCRow's lifetime does not allow it to exist after the function terminates.
+If the user wants to save the data for later 
+(e.g. in a consumer that saves all consumed rows in a `vec`)
+they should map the rows to another data structure.
+## Saving progress
+TODO
```
piodul, line 230: Please collaborate with @macher259 on this - I know that you are busy with the benchmarks right now, so maybe he can help you write this section if he has some free time to spare.

You can wait with this until https://github.com/piodul/scylla-cdc-rust/pull/69 is merged, or at least until you are fairly sure that the interface/semantics won't change further during the review.

tutorial.md
```diff
@@ -0,0 +1,231 @@
...
+        .end_timestamp(end)
+        .consumer_factory(factory)
+        .build()
+        .await
+        .expect("Creating the log reader failed!");
+
+    handle.await
+}
+```
+[//]: # (TODO: Add link to log reader builder's API documentation)
```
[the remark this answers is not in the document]
procrastinationfighter: Well, we haven't published the docs yet, so I'm not sure how to link to this now.

tutorial.md
```diff
@@ -0,0 +1,231 @@
...
+- to check if a column was deleted in given operation, use `is_value_deleted`
+- to check which elements were deleted from a collection, use `get_deleted_elements` or `take_deleted_elements`
+
+The data returned is of type [CqlValue](<https://docs.rs/scylla/latest/scylla/frame/response/result/enum.CqlValue.html>)
+from the driver.
+
+For more detailed information about how to interpret these values, 
+check the [CDC documentation](<https://docs.scylladb.com/using-scylla/cdc/cdc-log-table/>).
+
+[//]: # (TODO: Add link to the API documentation)
```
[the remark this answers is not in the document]
procrastinationfighter: Same as the comment about log reader's docs - I'm referring here to the crates docs that will be available once we've published the library. I'm not sure how to link to this now.

procrastinationfighter:
@macher259 has created the chapter about saving progress, I guess this PR is ready to be reviewed and merged.

procrastinationfighter:
I tried adding the links to the docs - it's not guaranteed that they'll work, but it seems fine. I also deleted the `installation` section from readme.

scylla-cdc/src/consumer.rs
```diff
@@ -351,8 +351,8 @@ mod tests {
             ],
             1,
         )
-        .await?
-        .0;
+            .await?
+            .0;
```
piodul: Unnecessary whitespace change

tutorial.md
```diff
@@ -0,0 +1,290 @@
...
+_________
+_________
+Time UUID: cc07e47a-d07c-11ec-ebc7-e1924abf3e16, operation type: RowUpdate
+pk: 1, ck: 2
+v: Some(8), was deleted: false
+vs: None, was deleted: false, deleted elements: []
+_________
+```
+
+<b>Note:</b> CDCRow's lifetime does not allow it to exist after the function terminates.
```
piodul, line 222: Nitpick: in Markdown, you can mark text as bold by putting it `__between double underscores__`.

scylla-cdc/src/consumer.rs
```diff
@@ -58,6 +74,8 @@ const TTL_NAME: &str = "cdc$ttl";
 const IS_DELETED_PREFIX: &str = "cdc$deleted_";
 const ARE_ELEMENTS_DELETED_PREFIX: &str = "cdc$deleted_elements_";
 
+/// An internal structure used to map names of columns in the CDC
+/// to their indices in an internal vector.
```
piodul, line 78: In what sense it is internal? It is a part of the public API.

tutorial.md
```diff
@@ -0,0 +1,290 @@
...
+There is a default implementation that uses ScyllaDB called `TableBackedCheckpointSaver`. 
+Its source code is located in file [checkpoints.rs](scylla-cdc/src/checkpoints.rs) and can serve as an inspiration how to write new checkpoint savers.
+
+Example of usage for `TableBackedCheckpointSaver`:
+```rust
+let user_checkpoint_saver = Arc::new(
+            TableBackedCheckpointSaver::new_with_default_ttl(session, "ks", "checkpoints")
+                .await
+                .unwrap(),
+        );
```
piodul, line 264: These lines are indented too much with compared to the first one. The last two snippets have the same problem, too.

scylla-cdc/src/log_reader.rs
```diff
@@ -1,3 +1,20 @@
+//! A module containing the logic behind reading and consuming the data.
+//!
+//! Only one stream generation is being read at a time,
+//! while data from each stream is being read concurrently by different StreamReaders.
+//!
+//! The reading is done by reading all CDC rows added in a time window
+//! (not longer than `window_size`), then waiting for amount of time equal to `sleep_interval`
+//! before reading the next window.
+//! Before a newly added row is being read by the reader,
+//! a set amount of time, represented by `safety_interval`, must pass.
```
piodul, line 10: Please rephrase this section so that it doesn't describe "the reading" but rather what a `StreamReader` is doing. Otherwise it might suggest that stream readers are synchronized in some way and they read a time window at the same time which isn't necessarily true.

tutorial.md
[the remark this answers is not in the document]
piodul: You already added some guessed links, so what about this line?

scylla-cdc/src/consumer.rs
procrastinationfighter: To be honest, it shouldn't be public. But we can't create the replication tests without it being public (we don't use the whole library there and we can't create `CDCRowSchema` if it's not public). I will just remove the adjective.

piodul:
Looks good, but there is one important thing missing. When I open the scylla-cdc's crate docs generated by `cargo doc`, I get a list of modules without any explanation:

image

This is because the crate level documentation is missing (no documentation for the main module of scylla-cdc). Some people might get to the docs.rs page skipping github.com and crates.io altogether and will miss the contents of README.md, so we definitely need to put something here (as many other crates do).

Please add documentation to the main module of scylla-cdc. This documentation doesn't have to be long and should tell the user:

- What the crate is about,
- Give a general idea about how to use the crate (e.g. a short example + links to the tutorial).

Please look at the main page of the [scylla-rust-driver](https://docs.rs/scylla/latest/scylla/) for inspiration.

procrastinationfighter:
Done. I also updated some docs, because they linked to the StreamReader, which is a hidden object now.

piodul (approved):

PR #92, opened 2022-05-18
Check for potentially missed generation

### LINKED ISSUE #72: Potentially missed generation
More information: https://github.com/piodul/scylla-cdc-rust/pull/10#discussion_r826865162
Location in the code: https://github.com/piodul/scylla-cdc-rust/blob/main/scylla-cdc/src/log_reader.rs#L196
I am creating this issue, because now it's only mentioned in the code and we could potentially forget about it.

Gor027:
Fixes #72
This PR adds additional checking for a potentially missed generation. As `next_generation`'s value is set to `None` during the execution and that logic is used in other places, I decided to not use `next_generation` as an argument to try to fetch the next generation, instead, a new variable `last_fetched_generation` is introduced to keep track of the last generation.

### DIFF
--- a/scylla-cdc/src/log_reader.rs
+++ b/scylla-cdc/src/log_reader.rs
@@ -68,7 +68,7 @@ impl CDCReaderWorker {
         let mut stream_reader_tasks = FuturesUnordered::new();
 
         let mut next_generation: Option<GenerationTimestamp> = None;
-        let mut had_first_generation: bool = false;
+        let mut current_generation: Option<GenerationTimestamp> = None;
         let mut err: Option<anyhow::Error> = None;
 
         loop {
@@ -88,7 +88,6 @@ impl CDCReaderWorker {
                 }
                 Some(generation) = generation_receiver.recv(), if next_generation.is_none() && err.is_none() => {
                     next_generation = Some(generation.clone());
-                    had_first_generation = true;
                     self.set_upper_timestamp(generation.timestamp).await;
                 }
                 Ok(_) = self.end_timestamp_receiver.changed(), if err.is_none() => {
@@ -104,6 +103,7 @@ impl CDCReaderWorker {
                 }
 
                 if let Some(generation) = next_generation.take() {
+                    current_generation = Some(generation.clone());
                     if generation.timestamp > self.end_timestamp {
                         return Ok(());
                     }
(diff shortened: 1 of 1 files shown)


scylla-cdc/src/log_reader.rs
```diff
@@ -141,10 +156,9 @@ impl CDCReaderWorker {
                             })
                         })
                         .collect();
-                } else if had_first_generation {
-                    // FIXME: There may be another generation coming in the future with timestamp < end_timestamp
-                    // that could have been missed because of earlier fetching failures.
-                    // More on this here https://github.com/piodul/scylla-cdc-rust/pull/10#discussion_r826865162
```
piodul, line 147: If we are here, then it means that all stream readers read up to the `end_timestamp` and stopped. The problem that may occur here is that a generation `G2` which is less than the `end_timestamp` appeared in scylla but the generation fetcher was delayed for some reason and we don't know about it yet. We need to check if `G2` exists - if it does, we need to start reading from it, otherwise we can stop.

The idea is to query the generations table here and see if such `G2` exists or not.

It looks like your solution attempts to solve this problem by waiting some time on the `generation_receiver` channel and it assumes that `G2` exists if it appears on the channel within some timeout. You cannot assume that if something doesn't happen within some timeout then it won't happen at all - in an asynchronous, non-real-time system there are many factors which may delay execution of code, e.g.:

- The runtime scheduler may have a very long queue of tasks and your code is at the end of it,
- The OS thread the application executes in may be preempted by other processes,
- A request to scylla takes long because the cluster is overloaded,

Etc.

----

The simplest way I can think of to fix the problem without using timeouts like that is:

1. Use `fetch_next_generation` here to see if there is a new generation which is lower than `end_timestamp`.
2. If there is, then continue the main loop. The generation fetcher will eventually notice this generation and will return it on the `generation_receiver` channel, then the log reader will read the range `G2..end_timestamp`.
3. If not, then quit.

scylla-cdc/src/log_reader.rs
```diff
@@ -141,10 +152,9 @@ impl CDCReaderWorker {
...
+                } else if !check_missed_generation_enabled {
+                    check_missed_generation_enabled = true;
```
piodul, line 156: Can't we call `fetcher.fetch_next_generation` here? Is it necessary to have the `check_missed_generation_enabled` flag and fo the fetch in the next loop iteration?

Adding more flags/variables further complicates the logic of `CDCReaderWorker::run` - which is already quite complicated - so it would be great if we could avoid it.

scylla-cdc/src/log_reader.rs
```diff
@@ -68,7 +68,7 @@ impl CDCReaderWorker {
         let mut stream_reader_tasks = FuturesUnordered::new();
 
         let mut next_generation: Option<GenerationTimestamp> = None;
-        let mut had_first_generation: bool = false;
+        let mut last_fetched_generation: Option<GenerationTimestamp> = None;
```
piodul, line 71: I think `current_generation` would be a more appropriate name. It's not the generation you have fetched most recently (next_generation fits this description better), but rather the one we are reading at the moment.

scylla-cdc/src/log_reader.rs
```diff
@@ -141,10 +141,25 @@ impl CDCReaderWorker {
...
+                            next_generation = Some(generation.clone());
+                            self.set_upper_timestamp(generation.timestamp).await;
+                            continue;
+                        } else {
+                            // Fetched next generation with bigger timestamp than end_timestamp
+                            return Ok(());
+                        }
+                    } else {
+                        // Last fetched generation's next generation is not available
+                        return Ok(());
```
piodul, line 159: We should at least print a warning here, describing what happened and warning about potentially unread rows.

piodul (approved):

PR #95, opened 2022-05-31
Optimization: make each StreamReader use multiple streams

macher259:
Previously, we mapped a StreamReader to each stream in 1-1 fashion. I have changed it to use multiple streams grouped by vnodes.


procrastinationfighter:
Worth noticing:
In #90 I have added some docstrings about one consumer being created for one stream id (in `consumer.rs`). In case this gets approved, this docstring should also be changed (I'm not sure which one gets merged first).

piodul (approved):

PR #98, opened 2022-06-08
Change modules visibility

procrastinationfighter:
Most modules present in `scylla-cdc` are not something that should have public visibility. This PR changes that, so that only `consumer` and `log_reader` modules are accessbile from outside the crate.

Because `test_utilities` module is used in the printer and the replicator, it couldn't simply become a hidden module. I used `cfg` and Cargo's `features` to make this module public only if the library is being used with a feature called `test`.

### DIFF
--- a/scylla-cdc-printer/Cargo.toml
+++ b/scylla-cdc-printer/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 
 [dependencies]
 scylla = { git = "https://github.com/scylladb/scylla-rust-driver" }
-scylla-cdc = { version = "0.1.0", path = "../scylla-cdc" }
+scylla-cdc = { version = "0.1.0", path = "../scylla-cdc", features = ["test"] }
 tokio = { version = "1.1.0", features = ["full"] }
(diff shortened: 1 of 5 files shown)


procrastinationfighter:
@piodul review, please.
I was mistaken when I was saying that only one function from that module is being used - actually all of them are important.

piodul (changes requested):
Let's use a "test ~package~ crate" pattern instead. The `test_utilities` package does not really depend on anything from `scylla-cdc` - it only uses `ToTimestamp` trait, but instead of using it you can inline the `to_timestamp` call.

piodul:
The main problem with using a feature is that we can't really forbid users from depending on it, so technically the test utilities become a part of the library's interface. When modifying the test utilities in a way that can potentially break somebody's code, it becomes unclear whether you need to bump the major version of the library or not. Using a dedicated crate should avoid this problem.

procrastinationfighter:
Since the required changes will change a major part of the pull request, I'm closing this PR and I'll open a new one.

PR #101, opened 2022-06-14
Change deprecated clap functions to their alternatives.

macher259:
Currently, in the printer and replicator, we implicitly use deprecated unit variant `clap::ArgAction::StoreValue`. This causes our workflows to fail, as we can see [here](https://github.com/piodul/scylla-cdc-rust/runs/6878180188?check_suite_focus=true#step:5:342). I have changed it to explicitly use `clap::ArgAction::Set` instead.

### DIFF
--- a/scylla-cdc-printer/src/main.rs
+++ b/scylla-cdc-printer/src/main.rs
@@ -12,27 +12,27 @@ use crate::printer::PrinterConsumerFactory;
 #[derive(Parser)]
 struct Args {
     /// Keyspace name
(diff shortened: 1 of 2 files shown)


piodul (approved):

piodul:
@macher259 as a followup, please bump the `clap` version in `Cargo.toml`. The `clap::ArgAction` feature you use here is only available since `3.2.0` and the code doesn't compile with older versions (I still had `3.1.18` in my `Cargo.lock` specified).

PR #103, opened 2022-06-18
Change modules visibility

procrastinationfighter:
The goal of this pull request is to hide some modules in the library. To do so, the module `test_utilities` has been moved to a separate crate (because it should not be public, but the replicator and the printer use it). 

Due to the saving progress interface introduced in #69, the `cdc_types` module should be made public. I am going to add documentation to this module in #90.


scylla-cdc-printer/Cargo.toml
```diff
@@ -14,3 +14,4 @@ futures = "0.3.17"
 anyhow = "1.0.51"
 async-trait = "0.1.52"
 clap = { version = "3.1.10", features = ["derive"] }
+scylla-cdc-utils = { path = "../scylla-cdc-utils" }
```
piodul, line 17: This should be put into the `dev-dependencies` section, as those utilities are only used in the tests.

scylla-cdc-replicator/Cargo.toml
```diff
@@ -18,3 +18,4 @@ futures-util = "0.3.19"
 tracing = "0.1.34"
 chrono = "0.4.19"
 clap = { version = "3.1.10", features = ["derive"] }
+scylla-cdc-utils = { path = "../scylla-cdc-utils" }
```
piodul, line 21: `dev-dependencies` here, too.

scylla-cdc/Cargo.toml
```diff
@@ -17,6 +17,7 @@ async-trait = "0.1.51"
 tracing = "0.1.31"
 itertools = "0.10.3"
 hex = "0.4.3"
+scylla-cdc-utils = { path = "../scylla-cdc-utils" }
```
piodul, line 20: `dev-dependencies`

scylla-cdc/src/stream_reader.rs
```diff
@@ -7,10 +7,11 @@ use futures::StreamExt;
 use itertools::{repeat_n, Itertools};
 use scylla::frame::response::result::CqlValue;
 use scylla::Session;
+use scylla_cdc_utils::ToTimestamp;
```
piodul, line 9: We are using `ToTimestamp` in just two places - one of them is in the non-test code due to which we need to specify `scylla-cdc-utils` in dependencies, not dev-dependencies. Can we get rid of it? It doesn't save much typing and I don't think it is worth keeping in these circumstances.

piodul (changes requested):

procrastinationfighter:
@piodul done, I think it's ready to merge.

scylla-cdc/src/stream_reader.rs
piodul: OK, so we _did_ use it in multiple places, I'm not sure why I thought otherwise this morning... Nevertheless, I think the switch to `now()` from utils is fine.

PR #108, opened 2022-06-29
Handle timeouts in StreamReader

### LINKED ISSUE #107: Queries in StreamReader should be retried on timeout
Currently, StreamReader will stop if it encounters an error when executing a query. One type of such an error is a timeout - it is not unusual for such type of an error to happen in real world scenarios (e.g. when a cluster is temporarily overloaded) and it is not fatal, so it should be handled. Let's add a simple retry mechanism - in case of an error, the query should be retried after `sleep_interval`.

An important note: the `execute_iter` method we are using returns an iterator over rows. As the rows are fetched page by page behind the scenes, an error may be happen in the middle of the iteration. When retrying a query, it is important to make sure that the rows already fed to the CDCConsumer are not fed by it for the second time. I suggest using the `execute_paged` interface and doing the paging ourselves.

procrastinationfighter:
Fixes #107 
This pull request changes the way the `StreamReader` reads the CDC log. Now, it uses `execute_paged` and sleeps in case of a timeout. The process panics in case of other database error than a timeout.

### DIFF
--- a/scylla-cdc/src/stream_reader.rs
+++ b/scylla-cdc/src/stream_reader.rs
@@ -4,16 +4,21 @@ use std::cmp::{max, min};
 use std::sync::Arc;
 use std::time;
 
-use futures::StreamExt;
+use itertools::Itertools;
 use scylla::frame::value::Timestamp;
 use scylla::prepared_statement::PreparedStatement;
+use scylla::transport::errors::{DbError, QueryError};
 use scylla::Session;
 use tokio::sync::watch;
 use tokio::time::sleep;
+use tracing::{enabled, warn};
 
 use crate::cdc_types::{GenerationTimestamp, StreamID};
 use crate::checkpoints::{start_saving_checkpoints, CDCCheckpointSaver, Checkpoint};
 use crate::consumer::{CDCRow, CDCRowSchema, Consumer};
+
+const BASIC_TIMEOUT_SLEEP_MS: u128 = 10;
+const TIMEOUT_FACTOR: u128 = 2;
 
 #[derive(Clone)]
 pub struct CDCReaderConfig {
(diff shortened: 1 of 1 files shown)


procrastinationfighter:
@piodul

scylla-cdc/src/stream_reader.rs
```diff
@@ -134,17 +139,44 @@ impl StreamReader {
...
+                        sleep(time::Duration::from_millis(sleep_after_timeout)).await;
+                        sleep_after_timeout = sleep_after_timeout * TIMEOUT_FACTOR;
+                        if sleep_after_timeout >= TIMEOUT_LIMIT {
+                            println!("WARN: The database keeps returning a timeout error. Trying again...");
+                            sleep_after_timeout = TIMEOUT_LIMIT;
+                        }
+
+                        next_state = state_clone;
+                    },
+                    Err(e) => panic!("The database returned the following error: {:?}", e),
```
piodul: We shouldn't panic here. This is not an unrecoverable error, we should gracefully stop and propagate the error. Please return it from the function.

Please also use the `Error::context` method to add some information about the source of the error, i.e. that it happened while trying to fetch rows from cdc logs (to differentiate this error from a user error).

scylla-cdc/src/stream_reader.rs
```diff
@@ -134,17 +139,44 @@ impl StreamReader {
...
+                                    .consume_cdc(CDCRow::from_row(row, &schema))
+                                    .await?;
+                            }
+                        }
+                        match x.paging_state {
+                            Some(state) => next_state = Some(state),
+                            None => break 'try_query,
+                        }
+                    },
+                    Err(QueryError::TimeoutError) => {
```
piodul: This error is for client side timeouts (AFAIK they are not implemented for regular queries yet, but it's OK if you handle this case explicitly now). You also need to handle server-side timeouts: https://github.com/scylladb/scylla-rust-driver/blob/9d402a92558485ee8343386c4204d9c1d6ca19c0/scylla/src/transport/errors.rs#L124

scylla-cdc/src/stream_reader.rs
```diff
@@ -202,6 +164,55 @@ impl StreamReader {
 
         Ok(())
     }
+
+    /// Reads the CDC log using exponential backoff to handle timeouts.
+    async fn read_log(
```
piodul, line 169: Nitpick: `read_log` is a very generic name, similar to `fetch_cdc`. Please give the function below a more descriptive name.

scylla-cdc/src/stream_reader.rs
```diff
@@ -4,17 +4,22 @@ use std::cmp::{max, min};
...
 use tokio::sync::watch;
 use tokio::time::sleep;
 
 use crate::cdc_types::{GenerationTimestamp, StreamID};
 use crate::checkpoints::{start_saving_checkpoints, CDCCheckpointSaver, Checkpoint};
 use crate::consumer::{CDCRow, CDCRowSchema, Consumer};
 
+const BASIC_TIMEOUT_SLEEP: u64 = 10;
+const TIMEOUT_FACTOR: u64 = 2;
+const TIMEOUT_LIMIT: u64 = 30_000;
```
piodul, line 21: Let's not hardcode the `TIMEOUT_LIMIT`. I think we can just use `sleep_interval` as a cap.

scylla-cdc/src/stream_reader.rs
```diff
@@ -134,17 +139,44 @@ impl StreamReader {
...
+                        match x.paging_state {
+                            Some(state) => next_state = Some(state),
+                            None => break 'try_query,
+                        }
+                    },
+                    Err(QueryError::TimeoutError) => {
+                        sleep(time::Duration::from_millis(sleep_after_timeout)).await;
+                        sleep_after_timeout = sleep_after_timeout * TIMEOUT_FACTOR;
+                        if sleep_after_timeout >= TIMEOUT_LIMIT {
+                            println!("WARN: The database keeps returning a timeout error. Trying again...");
```
piodul: Don't use `println!` to tell things to the user, use the `tracing` crate instead. This will let the users filter out messages as they please.

Moreover, I don't think that this message is very helpful. Let's put more information here. Instead, please print a message every time there was a timeout, and include information about which streams were queried, what is the current backoff time, what was the current query window (`window_begin..window_end`) and the page number.

scylla-cdc/src/stream_reader.rs
```diff
@@ -139,45 +139,7 @@ impl StreamReader {
...
-                            println!("WARN: The database keeps returning a timeout error. Trying again...");
-                            sleep_after_timeout = TIMEOUT_LIMIT;
-                        }
-
-                        next_state = state_clone;
-                    },
-                    Err(e) => panic!("The database returned the following error: {:?}", e),
-                }
-            }
+            self.read_log(&query_base, &values, &mut consumer).await?;
```
piodul, line 142: I'd recommend moving the logic out first, then add timeouts. This will make applying review suggestions easier (you won't have to fix the refactoring commit after modifying the other commit).

piodul (changes requested):
This PR needs a test. For example, you can introduce an interface which mocks the Session. You can add a mock which wraps the session and, returns a timeout several times and finally starts forwarding the `query_paged` calls to the real Scylla.

scylla-cdc/src/stream_reader.rs
procrastinationfighter: I wanted to to it the same way as [here](https://github.com/piodul/scylla-cdc-rust/blob/ae8ccb5f8d03b1177b55c89c2b91dbf4176cb827/scylla-cdc/src/log_reader.rs#L180). Should it be changed, too?

piodul: Yes, but in a separate PR.

procrastinationfighter:
I'm not sure what do you mean by creating an interface to mock the `Session`. Should I replace the `Arc<Session>` with some kind of `Arc<SessionWrapper>` in the `StreamReader`? This seems a bit weird.

procrastinationfighter:
Besides adding a test, all changes were applied.

scylla-cdc/src/stream_reader.rs
```diff
@@ -168,21 +172,71 @@ impl StreamReader {
...
+        if enabled!(tracing::Level::WARN) {
+            let ids_str = values[0..values.len() - 2]
+                .iter()
+                .map(|x| format!("0x{}", hex::encode(x.as_blob().unwrap())))
+                .join(", ");
+            let begin = values[values.len() - 2].as_duration().unwrap();
+            let end = values[values.len() - 1].as_duration().unwrap();
+
+            warn!(
+                "WARN: Timeout while fetching CDC rows for streams: {}.\n\
```
piodul, line 232: `WARN:` at the beginning of the string is redundant. The backend responsible for printing/storing messages logged with the `tracing` crate will usually include information about the verbosity.

procrastinationfighter: Quite funny - I was inspired by [this code](https://github.com/piodul/scylla-cdc-rust/blob/ae8ccb5f8d03b1177b55c89c2b91dbf4176cb827/scylla-cdc/src/checkpoints.rs#L42) from the repository.

piodul: Well, it slipped past my review. The code you pointed to should be fixed, too (and other instances of such problem, if there are any).

procrastinationfighter: I fixed this in #112

piodul: Even though you fixed the other place, here you still have a redundant `WARN` in the string :(

piodul (changes requested):
Needs rebase after recent merges.

scylla-cdc/src/stream_reader.rs
```diff
@@ -344,6 +375,41 @@ mod tests {
...
+            self.session.prepare(query).await
+        }
+
+        async fn execute_paged_statement(
+            &self,
+            statement: &PreparedStatement,
+            values: &[CqlValue],
+            paging_state: Option<Bytes>,
+        ) -> Result<QueryResult, QueryError> {
+            let mut num = self.counter.lock().await;
```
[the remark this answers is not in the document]
piodul: > By the way, what was the correct way to reduce the scope? The only solution I was able to invent was something like this:

I think that your solution is the nicest way to do it. Alternately, you can use `std::mem::drop(num)` to drop the mutex guard (`num`), but IMO introducing a block is more clear.

piodul (approved):

PR #110, opened 2022-06-30
Revert "Unpack stream_id_vec bindings in order to allow driver to calculate correct partition key

Ponewor:
This reverts commit ac071fde53f3b864a4f800773bf9a18b27d91634.


piodul (changes requested):
Please include the reason for the revert in the commit message.

piodul (approved):

Issue #96, opened 2022-06-01
Change driver's version to a fixed one


procrastinationfighter:
Currently, the library uses the Scylla Rust Driver by linking to the Github repository [(here)](https://github.com/piodul/scylla-cdc-rust/blob/main/scylla-cdc/Cargo.toml). Once the driver's release version has all needed features, we should change it to crates.io with a fixed version.

PR #111, opened 2022-06-30
Change required Scylla Rust Driver version to a fixed one.

macher259:
Fixes: #96


piodul (changes requested):
Two things about the commit message:

- Please use the present tense and the imperative form in the commit message subject, i.e. "Change required Scylla Rust Driver..."
- The commit message should explain why this particular commit is needed. Please add a short description to the commit message.

PR #112, opened 2022-07-02
Use the tracing crate to print the warning in log_reader.rs

procrastinationfighter:
This pull request changes the way the warning about potentially missed generation is printed.
Ref: https://github.com/piodul/scylla-cdc-rust/pull/108#discussion_r912328136


scylla-cdc/src/log_reader.rs
```diff
@@ -177,7 +178,7 @@ impl CDCReaderWorker {
                         }
                     } else {
                         // Current generation's next generation is not available
-                        println!("WARN: Next generation is not available, some rows in the CDC log table may be unread.");
+                        warn!("WARN: Next generation is not available, some rows in the CDC log table may be unread.");
```
piodul, line 181: The `WARN:` prefix in the string is now redundant.

piodul (approved):

PR #113, opened 2022-07-10
Log reader: Add test for handling errors

### LINKED ISSUE #100: Test for the error handling logic in CDCLogReader
Based on the recent benchmarks, there is a reason to believe that there is a bug in the error handling logic in the CDCLogReader. It needs to be debugged and test(s) should be added which verify that it works.

procrastinationfighter:
Fixes #100 

Also, see my comment about the source of the bug that provoked this issue: https://github.com/piodul/scylla-cdc-rust/issues/100#issuecomment-1179729824

### DIFF
--- a/scylla-cdc/src/log_reader.rs
+++ b/scylla-cdc/src/log_reader.rs
@@ -490,3 +490,59 @@ impl Default for CDCLogReaderBuilder {
         Self::new()
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


procrastinationfighter:
@piodul

procrastinationfighter:
As far as I know, @macher259 is doing this task, also covering more cases than this PR does. I'm closing this PR for now.

PR #126, opened 2025-01-06
bump scylla to 0.15.1

abastian:
Hi,

For my project I need to bump scylla driver to latest version (0.15.1 as of now), and since it also use scylla-cdc-rust, then I try to update it too...

Mainly just updating [deserialization mechanism](https://rust-driver.docs.scylladb.com/stable/migration-guides/0.15-deserialization.html) and replacing `query`/`execute` with `query_unpaged`, `execute_unpaged`, `query_iter` and `execute_single_page` (need reviews here, my adjustment might be incorrect)

### DIFF
--- a/scylla-cdc-printer/Cargo.toml
+++ b/scylla-cdc-printer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scylla-cdc-printer"
-version = "0.2.0"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/scylla-cdc-printer/Cargo.toml
+++ b/scylla-cdc-printer/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "scylla-cdc-printer"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-scylla = "0.13.1"
-scylla-cdc = { version = "0.2.0", path = "../scylla-cdc" }
+scylla = "0.15.1"
+scylla-cdc = { version = "0.3.0", path = "../scylla-cdc" }
 tokio = { version = "1.1.0", features = ["full"] }
 chrono = "0.4.19"
 futures = "0.3.17"
--- a/scylla-cdc-printer/Cargo.toml
+++ b/scylla-cdc-printer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scylla-cdc-printer"
-version = "0.1.2"
+version = "0.2.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/scylla-cdc-replicator/Cargo.toml
+++ b/scylla-cdc-replicator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scylla-cdc-replicator"
-version = "0.2.0"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/scylla-cdc-replicator/Cargo.toml
+++ b/scylla-cdc-replicator/Cargo.toml
@@ -1,18 +1,18 @@
 [package]
 name = "scylla-cdc-replicator"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-scylla = "0.13.1"
-scylla-cdc = { version = "0.2.0", path = "../scylla-cdc" }
+scylla = "0.15.1"
+scylla-cdc = { version = "0.3.0", path = "../scylla-cdc" }
 tokio = { version = "1.1.0", features = ["full"] }
 async-trait = "0.1.51"
 anyhow = "1.0.48"
 itertools = "0.13.0"
-uuid = {version = "1.0.0", features = ["v1"]}
+uuid = { version = "1.0.0", features = ["v1"] }
 futures-util = "0.3.19"
 tracing = "0.1.34"
 chrono = "0.4.19"
--- a/scylla-cdc-replicator/Cargo.toml
+++ b/scylla-cdc-replicator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scylla-cdc-replicator"
-version = "0.1.2"
+version = "0.2.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -76,7 +76,7 @@ mod tests {
         let schema = CDCRowSchema::new(result.column_specs());
 
         while let Some(log_res) = result.next().await {
-            let mut log = log_res.unwrap();
+            let mut log = log_res?;
             // handling udt specific, replacing src keyspace with dst keyspace
             for col_opt in log.columns.iter_mut().flatten() {
                 if let UserDefinedType { keyspace, .. } = col_opt {
--- a/scylla-cdc-replicator/src/replication_tests.rs
+++ b/scylla-cdc-replicator/src/replication_tests.rs
@@ -2,7 +2,7 @@
 mod tests {
     use crate::replicator_consumer::ReplicatorConsumer;
     use anyhow::anyhow;
-    use futures_util::FutureExt;
+    use futures_util::{FutureExt, StreamExt, TryStreamExt};
     use itertools::Itertools;
     use scylla::frame::response::result::CqlValue::{Boolean, Int, Text, UserDefinedType};
     use scylla::frame::response::result::{CqlValue, Row};
@@ -47,12 +47,13 @@ mod tests {
         name: &str,
         last_read: &mut (i64, i32),
     ) -> anyhow::Result<()> {
-        let result = session
-            .query(
+        let mut result = session
+            .query_iter(
                 format!("SELECT * FROM {}.{}_scylla_cdc_log", ks_src, name),
                 (),
             )
-            .await?;
+            .await?
+            .rows_stream::<Row>()?;
 
         let table_schema = session
             .get_cluster_data()
@@ -72,9 +73,10 @@ mod tests {
         )
         .await;
 
-        let schema = CDCRowSchema::new(&result.col_specs);
+        let schema = CDCRowSchema::new(result.column_specs());
 
-        for mut log in result.rows.unwrap_or_default() {
+        while let Some(log_res) = result.next().await {
+            let mut log = log_res.unwrap();
             // handling udt specific, replacing src keyspace with dst keyspace
             for col_opt in log.columns.iter_mut().flatten() {
                 if let UserDefinedType { keyspace, .. } = col_opt {
@@ -163,21 +165,23 @@ mod tests {
         name: &str,
     ) -> anyhow::Result<()> {
         let original_rows = session
-            .query(format!("SELECT * FROM {}.{}", ks_src, name), ())
+            .query_iter(format!("SELECT * FROM {}.{}", ks_src, name), ())
             .await?
-            .rows
-            .unwrap_or_default();
+            .rows_stream::<Row>()?
+            .try_collect::<Vec<_>>()
+            .await?;
         let replicated_rows = session
-            .query(format!("SELECT * FROM {}.{}", ks_dst, name), ())
+            .query_iter(format!("SELECT * FROM {}.{}", ks_dst, name), ())
             .await?
-            .rows
-            .unwrap_or_default();
+            .rows_stream::<Row>()?
+            .try_collect::<Vec<_>>()
+            .await?;
 
         if original_rows.len() == replicated_rows.len() {
             for (i, (original, replicated)) in
                 original_rows.iter().zip(replicated_rows.iter()).enumerate()
             {
-                if !equal_rows(original, replicated) {
+                if !equal_rows(&original, &replicated) {
                     fail_test(
                         name,
                         &original_rows,
@@ -207,7 +211,7 @@ mod tests {
         column_name: &str,
     ) -> anyhow::Result<(Vec<Row>, Vec<Row>)> {
         let original_rows = session
-            .query(
+            .query_iter(
                 format!(
                     "SELECT WRITETIME ({}) FROM {}.{}",
                     column_name, ks_src, table_name
@@ -215,10 +219,11 @@ mod tests {
                 (),
             )
             .await?
-            .rows
-            .unwrap_or_default();
+            .rows_stream::<Row>()?
+            .try_collect::<Vec<_>>()
+            .await?;
         let replicated_rows = session
-            .query(
+            .query_iter(
                 format!(
                     "SELECT WRITETIME ({}) FROM {}.{}",
                     column_name, ks_dst, table_name
@@ -226,8 +231,9 @@ mod tests {
                 (),
             )
             .await?
-            .rows
-            .unwrap_or_default();
+            .rows_stream::<Row>()?
+            .try_collect::<Vec<_>>()
+            .await?;
 
         Ok((original_rows, replicated_rows))
     }
@@ -327,7 +333,7 @@ mod tests {
         let mut last_read = (0, 0);
 
         for operation in operations {
-            session.query(operation, []).await?;
+            session.query_unpaged(operation, []).await?;
             replicate(
                 &session,
                 &ks_src,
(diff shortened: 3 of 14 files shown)


wprzytula:
Thanks for this contribution!

scylla-cdc-printer/Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-scylla = "0.13.1"
-scylla-cdc = { version = "0.2.0", path = "../scylla-cdc" }
+scylla = "0.15.1"
```
wprzytula, line 9: :wrench: As we're bumping a major version number of `scylla`, let's bump a major version number of dependant packages, too.
(`0.2.0` instead of `0.1.2`)

scylla-cdc-replicator/Cargo.toml
```diff
@@ -1,18 +1,18 @@
...
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-scylla = "0.13.1"
-scylla-cdc = { version = "0.2.0", path = "../scylla-cdc" }
+scylla = "0.15.1"
```
wprzytula, line 9: :wrench: ditto: major version bump is needed

scylla-cdc-replicator/src/replication_tests.rs
```diff
@@ -72,9 +73,10 @@ mod tests {
         )
         .await;
 
-        let schema = CDCRowSchema::new(&result.col_specs);
+        let schema = CDCRowSchema::new(result.column_specs());
 
-        for mut log in result.rows.unwrap_or_default() {
+        while let Some(log_res) = result.next().await {
+            let mut log = log_res.unwrap();
```
wprzytula, line 79: :wrench: Justify the `unwrap`. I'm quite sure we should return an error instead.

scylla-cdc-test-utils/Cargo.toml
```diff
@@ -1,11 +1,11 @@
...
+version = "0.1.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 anyhow = "1.0.48"
 chrono = "0.4.19"
-scylla = "0.13.1"
+scylla = "0.15.1"
```
wprzytula, line 11: :wrench: ditto: major version bump is needed

scylla-cdc/src/checkpoints.rs
```diff
@@ -187,7 +187,7 @@ impl CDCCheckpointSaver for TableBackedCheckpointSaver {
     async fn load_last_generation(&self) -> anyhow::Result<Option<GenerationTimestamp>> {
         let generation = self
             .session
-            .query(
+            .query_unpaged(
                 format!(
                     "SELECT generation FROM {}
                     WHERE stream_id = ?",
```
wprzytula, line 193: :thinking: This is a SELECT. I'm unsure if it's acceptable that we use `query_unpaged` for this. Is the query always expected to return at most one row? What about the tombstones encountered? @Lorak-mmk

scylla-cdc/src/checkpoints.rs
```diff
@@ -209,7 +210,7 @@ impl CDCCheckpointSaver for TableBackedCheckpointSaver {
     ) -> anyhow::Result<Option<chrono::Duration>> {
         Ok(self
             .session
-            .query(
+            .query_unpaged(
                 format!(
                     "SELECT time FROM {}
                     WHERE stream_id = ?
```
wprzytula, line 216: :thinking: ditto about SELECT with `query_unpaged`.

scylla-cdc/src/checkpoints.rs
```diff
@@ -252,12 +254,13 @@ mod tests {
...
         session
-            .query(format!("SELECT * FROM {}", table), ())
+            .query_unpaged(format!("SELECT * FROM {}", table), ())
             .await
             .unwrap()
-            .rows()
+            .into_rows_result()
+            .unwrap()
+            .rows::<(StreamID, GenerationTimestamp, value::CqlTimestamp)>()
             .unwrap()
```
wprzytula, line 263: :thinking: These are only tests, but still SELECTing a whole table as unpaged seems wrong.

scylla-cdc/src/consumer.rs
```diff
@@ -387,15 +388,17 @@ mod tests {
     async fn test_query() {
         let session = setup().await.unwrap();
         let result = session
-            .query(
+            .query_unpaged(
                 format!("SELECT * FROM {};", TEST_SINGLE_VALUE_CDC_TABLE),
                 (),
```
wprzytula, line 393: :thinking: ditto about unpaged SELECT

scylla-cdc/src/stream_generations.rs
```diff
@@ -96,11 +95,13 @@ impl GenerationFetcher {
...
+            .query_unpaged(query, (value::CqlTimestamp(time.num_milliseconds()),))
             .await?
-            .rows;
+            .into_rows_result()?
+            .maybe_first_row::<(GenerationTimestamp,)>()?
+            .map(|(ts,)| ts);
 
-        GenerationFetcher::return_single_row(result)
+        Ok(result)
     }
```
wprzytula, line 105: :thinking: This returns at most one row, but could possibly scan through many tombstones. @Lorak-mmk WDYT?

scylla-cdc/src/stream_generations.rs
```diff
@@ -128,9 +129,15 @@ impl GenerationFetcher {
...
+        let result = self
+            .session
+            .query_unpaged(query, (generation,))
+            .await?
+            .into_rows_result()?
+            .maybe_first_row::<(GenerationTimestamp,)>()?
+            .map(|(ts,)| ts);
 
-        GenerationFetcher::return_single_row(result)
+        Ok(result)
```
wprzytula, line 140: :thinking: ditto tombstones @Lorak-mmk

scylla-cdc/src/stream_generations.rs
```diff
@@ -238,15 +234,14 @@ async fn get_cluster_size(session: &Session) -> anyhow::Result<usize> {
...
-    let mut rows = session
-        .query("SELECT COUNT(*) FROM system.peers", &[])
+    let (rows,) = session
+        .query_unpaged("SELECT COUNT(*) FROM system.peers", &[])
         .await?
-        .rows
-        .unwrap()
-        .into_typed::<(i64,)>();
+        .into_rows_result()?
+        .first_row::<(i64,)>()?;
```
wprzytula, line 241: :thinking: Should `COUNT(*)` queries be done with paging? @Lorak-mmk

wprzytula: There shouldn't be many peers in the cluster ofc, so the current unpaged way is probably valid in this case, but just to be sure in a general case.

scylla-cdc/src/stream_generations.rs
```diff
@@ -238,15 +234,14 @@ async fn get_cluster_size(session: &Session) -> anyhow::Result<usize> {
...
 
     // Query returns a number of peers in a cluster, so we need to add 1 to count current node.
-    Ok(rows.next().unwrap().unwrap().0 as usize + 1)
+    Ok(rows as usize + 1)
```
wprzytula, line 244: :wrench: the variable name should be changed, because it's no more `rows` that is represented by the variable.
I suggest `peers_num`.

scylla-cdc/src/stream_reader.rs
```diff
@@ -43,8 +45,8 @@ trait StreamSession: Sync + Send {
         ids: &[StreamID],
         window_begin: &value::CqlTimestamp,
         window_end: &value::CqlTimestamp,
-        paging_state: Option<Bytes>,
-    ) -> Result<QueryResult, QueryError>;
+        paging_state: PagingState,
+    ) -> Result<(QueryResult, Option<PagingState>), QueryError>;
```
wprzytula, line 49: :wrench: I suggest returning `PagingStateResponse` here rather than `Option<PagingState>`, to be more in line with the abstract meaning of both.

scylla-cdc/src/stream_reader.rs
```diff
@@ -407,8 +417,8 @@ mod tests {
             ids: &[StreamID],
             window_begin: &value::CqlTimestamp,
             window_end: &value::CqlTimestamp,
-            paging_state: Option<Bytes>,
-        ) -> Result<QueryResult, QueryError> {
+            paging_state: PagingState,
+        ) -> Result<(QueryResult, Option<PagingState>), QueryError> {
```
wprzytula, line 421: :wrench: ditto about returning `PagingStateResponse`.

wprzytula (changes requested):
:100: Really good job!
Mainly minor comments.

On our (maintainers') side, we still have to consider if making unpaged SELECT queries in certain cases is OK or not.

scylla-cdc-replicator/src/replication_tests.rs
abastian: ah... ic... got blindside since previous use `unwrap` variant (`unwrap_or_default()` exactly), but now you mention it, this `unwrap_or_default()` will not cause panic...

scylla-cdc/src/checkpoints.rs
abastian: ic... I'll use `query_iter()` for this

scylla-cdc/src/consumer.rs
abastian: same as above, I'll use `query_iter` here too

scylla-cdc-printer/Cargo.toml
Lorak-mmk: `scylla-cdc-printer` is a command line app, version bumps of `scylla` don't affect it that way.
For a command line app the closes thing we have to public API is its interface (arguments, flags etc). If those don't change, then I see no reason to bump the major version.
The same is true for `scylla-cdc-replicator`.

scylla-cdc/src/checkpoints.rs
Lorak-mmk: I'm don't know how this works with cdc tables (are there tombstones there?). @piodul

scylla-cdc/src/stream_generations.rs
Lorak-mmk: Again, no idea. cc: @piodul

scylla-cdc/src/stream_generations.rs
wprzytula: > 🤔 Should `COUNT(*)` queries be done with paging? @Lorak-mmk

I've got an answer: they should.

abastian: hm? Isn't this query always return one row?

wprzytula: It's, unfortunately, not as simple as the number of rows returned...
IIUC, it's possible that ScyllaDB returns empty pages when executing a filtered SELECT.
What's the issue in this particular case though? Perhaps @fee-mendes can explain this.

Sh4d1:
Hey! Nice work 😄  We were looking to bump our scylla driver (and CDC) and found this 😄 @wprzytula have you (the maintainers) decided on the paged/unpaged queries ? If I'm not mistaken on the current version it's only doing unpaged queries ? So wouldn't be any change there, and leave time for more discussion about that?

Lorak-mmk:
Imo lets be safe and avoid any unpaged queries.

scylla-cdc/src/checkpoints.rs
piodul: Why is it a concern here? This PR only bumps up the driver version, let's not use it to fix bugs if there are any. `query` was renamed to `query_unpaged`, so let's do only that here.

Besides, look at the schema of the checkpoint table. It only has a partition key (`stream_id`) and no clustering key. This query retrieves a single partition, which will be a single row.

scylla-cdc/src/checkpoints.rs
piodul: Again, this is fine.

scylla-cdc/src/consumer.rs
piodul: This is only a test, it's fine to use an unpaged query here.

scylla-cdc/src/checkpoints.rs
Lorak-mmk: This convinces me, changing query types can be done at another time (if at all).

scylla-cdc/src/stream_generations.rs
piodul: It's fine, the table being scanned here is not big and rarely changes (only during topology changes). The tombstones should expire quickly enough that they do not cause performance problems.

scylla-cdc/src/stream_generations.rs
piodul: Again, not a problem.

scylla-cdc/src/stream_generations.rs
piodul: Although this query does a scan, the scan is done on a local table and the table is small, so it should be rather quick.

abastian: hmm... I'll use `query_iter()` here too, better safe as @Lorak-mmk mention

Lorak-mmk:
> Imo lets be safe and avoid any unpaged queries.

@abastian Sorry for the confusion. Please see @piodul 's comments in the review. He responded to all of them regarding paged vs unpaged, with convincing arguments. Please apply his comments.

scylla-cdc/src/stream_generations.rs
abastian: ah... ic... i have already committing `query_iter()`, I revert it then

scylla-cdc/src/checkpoints.rs
wprzytula: > Why is it a concern here? This PR only bumps up the driver version, let's not use it to fix bugs if there are any.

I slightly disagree. The point of the changes introduced in 0.14 (`query` -> `query_unpaged`/`query_single_page`) was to require users to make conscious decision wrt paging, preferably straight upon migration. In this library we are users of the driver and thus we should reconsider our use of paging methods right upon migration.

wprzytula:
@abastian Please let us know once you finish the requested changes.

abastian:
> @abastian Please let us know once you finish the requested changes.

@wprzytula, I have finish the requested changes, please re-review, I hope these can be resolved...

wprzytula (approved):

wprzytula:
@Lorak-mmk Please take a look and merge once ready.

scylla-cdc/src/cdc_types.rs
```diff
@@ -29,15 +29,20 @@ impl SerializeCql for GenerationTimestamp {
...
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<scylla::deserialize::FrameSlice<'frame>>,
+    ) -> Result<Self, scylla::deserialize::DeserializationError> {
+        let timestamp = <CqlTimestamp as DeserializeValue<'frame, 'metadata>>::deserialize(typ, v)?;
+
+        Ok(GenerationTimestamp {
+            timestamp: chrono::Duration::milliseconds(timestamp.0),
+        })
```
Lorak-mmk, line 45: I'm curious, do you know if there is a reason GenerationTimestamp stores `timestamp` as `chrono::Duration` instead of `DateTime` / `NaiveDateTime`? `Duration` should be used to represent offset between two points in time, not a single point in time.

It would also simplify this a bit, because `DateTime` / `NaiveDateTime` are supported by Rust Driver.

scylla-cdc/src/cdc_types.rs
```diff
@@ -64,12 +69,16 @@ impl SerializeCql for StreamID {
...
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for StreamID {
+    fn type_check(typ: &ColumnType) -> Result<(), scylla::deserialize::TypeCheckError> {
+        <Vec<u8> as DeserializeValue<'frame, 'metadata>>::type_check(typ)
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<scylla::deserialize::FrameSlice<'frame>>,
+    ) -> Result<Self, scylla::deserialize::DeserializationError> {
+        let id = <Vec<u8> as DeserializeValue<'frame, 'metadata>>::deserialize(typ, v)?;
```
Lorak-mmk, line 81: @wprzytula maybe we should have an attribute for our derive macros, let's call it `transparent` that could be used for single-field structs / enum structs? It is a common pattern I see when people use the driver that they have to manually implement the traits in such case.

Lorak-mmk (approved):
Note to whoever will be merging this: squash merge would be the most appropriate here, because of the unoptimal commit structure (later commits fix changes introduce in earlier commits)

scylla-cdc/src/cdc_types.rs
wprzytula: :100: this is a great idea!

wprzytula:
:rocket: Thank you for the contribution, @abastian!

PR #127, opened 2025-03-26
Update Scylla Rust Driver to 1.0

Lorak-mmk:
Most of the changes were trivial - different import paths, different names of structs and methods, and some differences in errors.

The only non-obvious change is caused by ColumnType becoming non_exhaustive. I had to handle "_" case in "update_or_insert" because of that, for which I had to introduce an error type. This is in scylla-cdc-replicator, so has no influence on public API.

While doing that I adeed Vector type handling in the same function.


Lorak-mmk:
I am unable to request review using Github, so just tagging here: @wprzytula @muzarski @piodul

Lorak-mmk:
Oops, forgot about doctests.

muzarski (approved):

wprzytula (approved):

wprzytula:
> Would you like to do this or should I?

I'll happily do the release.

PR #145, opened 2026-01-16
General fixes

adespawn:
This PR improves the handling of the database timeouts and errors. This includes:

- marking all select statements idempotent to allow for driver retry policy retries. 
- improving logging, when a timeout  that we ignore occurs
- improving backoff logic to increase type safety and set more reasonable starting values

While this does not directly fixes the DRIVER-7, the goal here is to improve the behaviour in those kind of situations

### DIFF
--- a/scylla-cdc/src/stream_reader.rs
+++ b/scylla-cdc/src/stream_reader.rs
@@ -21,7 +21,7 @@ use crate::cdc_types::{GenerationTimestamp, StreamID};
 use crate::checkpoints::{CDCCheckpointSaver, Checkpoint, start_saving_checkpoints};
 use crate::consumer::{CDCRow, CDCRowSchema, Consumer};
 
-const BASIC_TIMEOUT_SLEEP: tokio::time::Duration = tokio::time::Duration::from_millis(10);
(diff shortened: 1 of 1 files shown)


scylla-cdc/src/stream_generations.rs
```diff
@@ -442,8 +448,11 @@ async fn get_cluster_size(session: &Session) -> anyhow::Result<usize> {
     // We are using default consistency here since the system keyspace is special and
     // the coordinator which handles the query will only read local data
     // and will not contact other nodes, so the query will work with any cluster size larger than 0.
+    let mut statement = Statement::from("SELECT COUNT(*) FROM system.peers");
+    statement.set_is_idempotent(true);
```
Lorak-mmk: Why did you not use the util function here?

scylla-cdc/src/stream_reader.rs
Lorak-mmk: Commit: `Update timeout warning to include driver error`

It looks like most of the changes in `scylla-cdc/src/stream_reader.rs` are unrelated to what is described in the commit message. If you performed them intentionally, could you move them to a separate commit?

wprzytula: The better way is to employ the `@` operator, if it's possible.

wprzytula (changes requested):

PR #150, opened 2026-01-20
treewide: replace `chrono::Duration` with `std::time::Duration`

### LINKED ISSUE #147: Inconsistency with CDCCheckpointSaver trait using Duration
The `Checkpoint` struct uses `core::time::Duration` and `CDCCheckpointSaver` uses it in a `save_checkpoint` function. But the trait uses also `chrono::Duration` in a `load_last_checkpoint` function. This is inconvenient inconsistency (I need to depend on `chrono` crate and use conversion internally.

adespawn:
Before this change, parts of the public API used `chrono::Duration`
while others used `std::time::Duration`, with no good reason for the
split. This inconsistency was reported by internal user (see #147).

Replace `chrono::Duration` with `std::time::Duration` throughout the
public API and all internal code, and remove the chrono dependency from
`scylla-cdc` and `scylla-cdc-test-utils`.

`chrono` is still used in a few places to retrieve the current time.

Fixes: #147
Fixes: VECTOR-729

### DIFF
--- a/scylla-cdc/src/cdc_types.rs
+++ b/scylla-cdc/src/cdc_types.rs
@@ -26,7 +26,6 @@ use std::time::UNIX_EPOCH;
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 pub(crate) struct Timestamp(i64);
 
-#[allow(dead_code)]
 impl Timestamp {
     /// The earliest representable `Timestamp` (milliseconds = `i64::MIN`).
     pub(crate) const MIN: Self = Timestamp(i64::MIN);
@@ -174,12 +173,12 @@ impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for Timestamp {
 /// A struct representing a timestamp of a stream generation.
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct GenerationTimestamp {
-    pub(crate) timestamp: chrono::Duration,
+    pub(crate) timestamp: Timestamp,
 }
 
 impl fmt::Display for GenerationTimestamp {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.timestamp.num_milliseconds())
+        write!(f, "{}", self.timestamp.as_millis())
     }
 }
 
@@ -189,23 +188,21 @@ impl SerializeValue for GenerationTimestamp {
         typ: &ColumnType,
         writer: CellWriter<'b>,
     ) -> Result<WrittenCellProof<'b>, SerializationError> {
-        CqlTimestamp(self.timestamp.num_milliseconds()).serialize(typ, writer)
+        self.timestamp.serialize(typ, writer)
     }
 }
 
 impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for GenerationTimestamp {
     fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
-        <CqlTimestamp as DeserializeValue<'frame, 'metadata>>::type_check(typ)
+        Timestamp::type_check(typ)
     }
 
     fn deserialize(
         typ: &'metadata ColumnType<'metadata>,
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError> {
-        let timestamp = <CqlTimestamp as DeserializeValue<'frame, 'metadata>>::deserialize(typ, v)?;
-
         Ok(GenerationTimestamp {
-            timestamp: chrono::Duration::milliseconds(timestamp.0),
+            timestamp: Timestamp::deserialize(typ, v)?,
         })
     }
 }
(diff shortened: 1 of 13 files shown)


wprzytula (approved):

wprzytula (changes requested):
Actually, Copilot pointed out important things.

QuerthDP:
@adespawn I'll take this PR since https://github.com/scylladb/scylla-cdc-rust/pull/162 will expose a new "last checkpoint" value which should be one of `chrono::Duration` / `std::time::Duration` so let's unify to the latter.

QuerthDP:
Changelog:
- rebased on main

QuerthDP:
Changelog:
- rebased on main

scylla-cdc/src/cdc_types.rs
```diff
@@ -8,6 +11,107 @@ use scylla::serialize::writers::WrittenCellProof;
...
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        let ts = <CqlTimestamp as DeserializeValue<'frame, 'metadata>>::deserialize(typ, v)?;
+        Ok(Timestamp(ts.0))
+    }
+}
```
wprzytula, line 171: ⚠️ Looks like a clone of Rust Driver's `CqlTimestamp`.

QuerthDP: `CqlTimestamp` is a raw wire type - it's a thin newtype in the Scylla driver with no behavior beyond CQL serialization/deserialization. `Timestamp` is a domain-layer wrapper built on top of it.

Specifically, `CqlTimestamp` lacks:
1. `now()` - a way to get the current wall-clock time
2. Arithmetic - `Timestamp` implements `Add<Duration>` and `Sub<Duration>` (with saturating arithmetic to prevent overflow), and `Sub<Timestamp> -> Duration` to compute intervals between instants
3. Sentinels - `Timestamp::MIN` / `Timestamp::MAX` constants

Timestamp still delegates to `CqlTimestamp` for the actual wire encoding, so it's not a replacement - it's a domain wrapper that uses `CqlTimestamp` only as a serialization primitive.

wprzytula: @Lorak-mmk Shall we bring such functionality to Rust Driver's `CqlTimestamp` or keep it on a separate wrapper?

QuerthDP: Whether you need it in Rust Driver or not, I think we could proceed with that here, and eventually remove that when it's available in the Rust Driver.

wprzytula: Agreed.
The comments I and Karol wrote in the PR to Rust Driver apply here, too. Let's continue the discussion there.

QuerthDP: https://github.com/scylladb/scylla-rust-driver/pull/1776 merged, I'll refactor `Timestamp` to match the adjusted `CqlTimestamp` so that the future substitution (once Rust Driver 1.8.0 is released) will be close to painless.

QuerthDP:
Changelog:
- updated `Timestamp` to match Rust Driver's `CqlTimestamp` adjusted in https://github.com/scylladb/scylla-rust-driver/pull/1776

wprzytula (approved):

scylla-cdc/src/cdc_types.rs
```diff
@@ -8,16 +11,174 @@ use scylla::serialize::writers::WrittenCellProof;
...
+
+    /// # Panics
+    ///
+    /// Panics if the resulting timestamp would overflow `i64` milliseconds.
+    fn add(self, rhs: Duration) -> Timestamp {
+        let rhs_ms = i128::try_from(rhs.as_millis()).unwrap_or(i128::MAX);
+        let result = i128::from(self.0) + rhs_ms;
+        Timestamp(i64::try_from(result).expect("overflow when adding Duration to Timestamp"))
+    }
+}
```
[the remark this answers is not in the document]
QuerthDP: This is an intentional design decision mirroring the upstream driver's `CqlTimestamp`, which went through the same saturating-vs-panicking discussion: https://github.com/scylladb/scylla-rust-driver/pull/1776

scylla-cdc/src/cdc_types.rs
```diff
@@ -8,16 +11,174 @@ use scylla::serialize::writers::WrittenCellProof;
...
+
+    /// # Panics
+    ///
+    /// Panics if the resulting timestamp would overflow `i64` milliseconds.
+    fn sub(self, rhs: Duration) -> Timestamp {
+        let rhs_ms = i128::try_from(rhs.as_millis()).unwrap_or(i128::MAX);
+        let result = i128::from(self.0) - rhs_ms;
+        Timestamp(i64::try_from(result).expect("overflow when subtracting Duration from Timestamp"))
+    }
+}
```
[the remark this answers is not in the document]
QuerthDP: Same as above: https://github.com/scylladb/scylla-rust-driver/pull/1776

PR #156, opened 2026-03-19
Introduce CqlIdentifier type to quote keyspace/table names in CQL

QuerthDP:
Add a CqlIdentifier newtype that wraps identifier names and always produces properly double-quoted output via its Display impl. This prevents CQL from case-folding names, which broke keyspaces with mixed-case names (e.g. those created by Alternator).

All CQL query construction sites now use CqlIdentifier instead of bare string interpolation:
- stream_reader: SELECT from CDC log table
- checkpoints: CREATE TABLE / UPDATE for checkpoint tables
- replicator_consumer: INSERT / DELETE / UPDATE queries (was previously unquoted entirely, which is a correctness bug)

Embedded double-quote characters are escaped by doubling them, following the CQL grammar.

Fixes: [VECTOR-584](https://scylladb.atlassian.net/browse/VECTOR-584)

[VECTOR-584]: https://scylladb.atlassian.net/browse/VECTOR-584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scylla-cdc/src/cql_identifier.rs
+++ b/scylla-cdc/src/cql_identifier.rs
@@ -1,0 +1,175 @@
+/// A CQL identifier (keyspace name, table name, column name, etc.) that is
+/// always properly quoted when formatted for use in CQL statements.
+///
+/// The inner value stores the name exactly as provided.
+/// The [`Display`](std::fmt::Display) implementation wraps it in double quotes
+/// with embedded double-quote characters escaped by doubling them
+/// (`"` -> `""`), following the CQL grammar.
+///
+/// Names coming from the Scylla driver's cluster metadata are already in
+/// their canonical internal form, so they should be passed through as-is.
+/// The quoting ensures CQL does not case-fold or reject reserved words.
+#[derive(Clone, Debug)]
+pub struct CqlIdentifier {
+    raw: String,
+    quoted: String,
+}
+
+impl CqlIdentifier {
+    /// Creates a new `CqlIdentifier`, preserving the name exactly as given.
+    ///
+    /// The name will be double-quoted when formatted, preventing CQL from case-folding it.
+    /// Pass names in their canonical form (e.g. as read from cluster metadata).
+    pub fn new(name: impl Into<String>) -> Self {
+        let raw = name.into();
+        let quoted = format!(r#""{}""#, raw.replace('"', r#""""#));
+        CqlIdentifier { raw, quoted }
+    }
+
+    /// Returns the raw (unquoted) identifier name.
+    pub fn as_raw(&self) -> &str {
+        &self.raw
+    }
+}
+
+impl PartialEq for CqlIdentifier {
+    fn eq(&self, other: &Self) -> bool {
+        self.raw == other.raw
+    }
+}
+
+impl Eq for CqlIdentifier {}
+
+impl std::hash::Hash for CqlIdentifier {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.raw.hash(state);
+    }
+}
+
+impl std::fmt::Display for CqlIdentifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.quoted)
+    }
+}
+
+impl From<&str> for CqlIdentifier {
+    fn from(s: &str) -> Self {
+        CqlIdentifier::new(s)
+    }
+}
+
+impl From<String> for CqlIdentifier {
+    fn from(s: String) -> Self {
+        CqlIdentifier::new(s)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::CqlIdentifier;
+
+    #[test]
+    fn quotes_lowercase_name() {
+        let id = CqlIdentifier::new("my_table");
+        assert_eq!(id.to_string(), r#""my_table""#);
+    }
+
+    #[test]
+    fn preserves_mixed_case() {
+        let id = CqlIdentifier::new("MyKeyspace");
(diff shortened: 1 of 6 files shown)


QuerthDP:
Changelog:
- move `CqlIdentifier` to separate module
- adjust tests

QuerthDP:
Changelog:
- add special characters test

QuerthDP:
@wprzytula @adespawn please take a look

QuerthDP:
Changelog:
- fix case-insensitive tests using uppercase table names :)

scylla-cdc/src/cql_identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        &self.0
+    }
+}
+
+impl std::fmt::Display for CqlIdentifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "\"{}\"", self.0.replace('"', "\"\""))
+    }
+}
+
```
Lorak-mmk, line 54: This `.replace` call will replace `"` with `""` in the whole name. What is the purpose of that? Is the double quote how you escape a quote in cql?

QuerthDP: According to [Scylla Docs](https://docs.scylladb.com/manual/stable/cql/definitions.html#identifiers):
> Inside a quoted identifier, the double-quote character can be repeated to escape it, so "foo "" bar" is a valid identifier.

Lorak-mmk (approved):

scylla-cdc/src/cql_identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn as_raw(&self) -> &str {
+        &self.0
+    }
+}
+
+impl std::fmt::Display for CqlIdentifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "\"{}\"", self.0.replace('"', "\"\""))
+    }
+}
```
wprzytula, line 53: ⛏️ Performance nit: isn't it more reasonable to allocate the quoted string on `CqlIdentifier` construction, not on each `Display::fmt` call?

scylla-cdc/src/cql_identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        let id = CqlIdentifier::new("My_Keyspace");
+        assert_eq!(id.as_raw(), "My_Keyspace");
+    }
+
+    #[test]
+    fn as_raw_roundtrips_with_new() {
+        let name = "some_name";
+        let id = CqlIdentifier::new(name);
+        assert_eq!(id.as_raw(), name);
+    }
```
wprzytula: ❓ What's the purpose of the second test??

scylla-cdc/src/cql_identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        let id = CqlIdentifier::new("  spaces  ");
+        assert_eq!(id.to_string(), "\"  spaces  \"");
+        assert_eq!(id.as_raw(), "  spaces  ");
+    }
+
+    #[test]
+    fn reserved_word_is_safely_quoted() {
+        let id = CqlIdentifier::new("select");
+        assert_eq!(id.to_string(), "\"select\"");
+    }
```
wprzytula: ❓ Any word will be safely quoted, right? It's nothing special about this for reserved words specifically.

scylla-cdc/src/cql_identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        let id = CqlIdentifier::new("col$name.with-special!chars");
+        assert_eq!(id.to_string(), "\"col$name.with-special!chars\"");
+        assert_eq!(id.as_raw(), "col$name.with-special!chars");
+    }
+
+    #[test]
+    fn display_composes_into_dotted_keyspace_table() {
+        let ks = CqlIdentifier::new("MyKeyspace");
+        let tbl = CqlIdentifier::new("my_table");
+        assert_eq!(format!("{ks}.{tbl}"), "\"MyKeyspace\".\"my_table\"");
```
wprzytula: ⛏️ Can you use raw strings to avoid escaping quotes using backslashes?

wprzytula (changes requested):

scylla-cdc/src/cql_identifier.rs
QuerthDP: It was redundant, `as_raw_roundtrips_with_new` tested the same property as `as_raw_returns_original_name`. Removed it.

scylla-cdc/src/cql_identifier.rs
QuerthDP: Correct. All identifiers are unconditionally quoted, so there's nothing special about reserved words. The test was misleading and has been removed.

scylla-cdc/src/cql_identifier.rs
QuerthDP: Changed

scylla-cdc/src/cql_identifier.rs
QuerthDP: Changed

QuerthDP:
Changelog:

- pre-compute quoted string at construction time instead of on each Display::fmt call
- remove redundant as_raw_roundtrips_with_new and reserved_word_is_safely_quoted tests
- use raw strings (r#"..."#) in all test assertions

wprzytula (approved):
Thanks!

PR #158, opened 2026-03-26
scylla-cdc: add internal operation types used by Alternator

QuerthDP:
Add two new `OperationType` variants with negative discriminants (-3, -4) used internally by Alternator to represent partition and row deletions in the CDC log. Explicit discriminant values are also assigned to all existing variants for clarity.

Add an `is_expiration` flag to CDCRow that indicates whether the operation was caused by a TTL expiration (i.e. the raw operation value is negative).

Fixes: VECTOR-588


QuerthDP:
Changelog:
- added `is_expiration` field to inform if the deletion came from user request or TTL expiration
- removed the new `OperationType` variants as that was breaking change, instead we use alternative on existing variants

QuerthDP:
Changelog:
- update commit message

Lorak-mmk (approved):

PR #162, opened 2026-07-06
[DO NOT REVIEW] scylla-cdc: expose consumed-window high-water mark via watch channel

QuerthDP:
# [WIP] [DO NOT REVIEW]

Add a per-stream progress reporting path and aggregate it into a global tokio::sync::watch channel accessible on CDCLogReader.

stream_reader.rs
- New pub(crate) trait StreamProgress with a single report() method. Implementors receive the fully-consumed window boundary (window_begin of the next window) as a chrono::Duration unix timestamp after every completed window and once more on clean exit.
- CDCReaderConfig gains pub(crate) window_progress: Option<Arc<dyn StreamProgress>>.
- fetch_cdc publishes the boundary at the normal advance site (lines 333-335) and again before break on clean exit so the last partial window is never silently dropped.

log_reader.rs
- GenerationProgress holds one chrono::Duration slot per reader (sentinel chrono::Duration::MIN) plus a Notify that wakes the aggregation task on any update.
- PerStreamProgressReporter implements StreamProgress and writes to its assigned slot, then fires the Notify.
- run_progress_aggregator: background task that loops on notify.notified(), computes min(positions), and sends to the global watch sender.  Runs until aborted.
- CDCReaderWorker gains window_progress_tx and aggregation_handle. On every generation transition the previous task is aborted, the global watch is reset to MIN (sentinel), a fresh GenerationProgress is built with the new reader count, and each reader is wired up with its own PerStreamProgressReporter slot.
- impl Drop for CDCReaderWorker aborts the aggregation task on exit.
- CDCLogReader gains window_progress_rx and a window_progress() accessor that returns a cloned Receiver<chrono::Duration>.
- CDCLogReaderBuilder::build creates the watch::channel(MIN) pair and threads it through.

Semantics: value is unix-time of min(consumed window boundary) across all active streams; advances toward now-safety_interval in steady state; chrono::Duration::MIN is the sentinel until the first window completes and on every generation transition.

Tests
- test_window_progress_initial_sentinel: asserts MIN immediately after build() (guards pre-first-window / empty-readers case).
- test_window_progress_advances: waits up to 30 s for the receiver to move past MIN using a 100 ms safety/sleep interval; asserts the final value is >= start_timestamp.

Fixes: VECTOR-727

### DIFF
--- a/scylla-cdc/src/log_reader.rs
+++ b/scylla-cdc/src/log_reader.rs
@@ -14,8 +14,7 @@
 
 use std::cmp::max;
 use std::sync::Arc;
-use std::time;
-use std::time::SystemTime;
+use std::time::{self, Duration, SystemTime};
 
 use anyhow;
 use futures::FutureExt;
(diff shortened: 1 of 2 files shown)


QuerthDP:
Closing as explained in: https://scylladb.atlassian.net/browse/VECTOR-727?focusedCommentId=76459

PR #164, opened 2026-07-07
treewide: format imports to one per line

QuerthDP:
Split multi-item imports across all source files to improve code readability and consistency. Each import statement now occupies its own line, following Rust formatting best practices.


Lorak-mmk:
How did you do this, and how do you plan to enforce that in the future? 

Afaik doing that requires a nightly rustfmt, which would be a pain to use by comtributors and CI.

QuerthDP:
> How did you do this, and how do you plan to enforce that in the future?

I did it semi-manually (Claude Haiku). The enforcement would be the maintainers requests to keep that style.

> Afaik doing that requires a nightly rustfmt, which would be a pain to use by comtributors and CI.

I don't want to enforce it with CI (I would but I'm afraid it's not straightforward as you said).
All the patch is is a simple style change, to improve readability in this project (which is not the best for now) so that current contributors would benefit from it.

QuerthDP:
## Motivation and value:

### Readability
 
When multiple items are imported on a single line, it's harder to scan what a module depends on at a glance. One import per line makes each dependency immediately visible without parsing brace-delimited lists.

### Cleaner diffs

Adding or removing a single import in a grouped statement touches a line shared with other items, creating noisy diffs and potential merge conflicts. With one-per-line, any change to imports is a clean single-line addition or deletion - the diff precisely reflects what changed.

### Consistency

The codebase currently has mixed style (some grouped, some flat). This patch establishes a uniform convention, reducing cognitive overhead when reading across different files.

@wprzytula please consider approving

piodul:
> > How did you do this, and how do you plan to enforce that in the future?
> 
> I did it semi-manually (Claude Haiku). The enforcement would be the maintainers requests to keep that style.
> 
> > Afaik doing that requires a nightly rustfmt, which would be a pain to use by comtributors and CI.
> 
> I don't want to enforce it with CI (I would but I'm afraid it's not straightforward as you said). All the patch is is a simple style change, to improve readability in this project (which is not the best for now) so that current contributors would benefit from it.

Readability of the imports is a rather subjective thing. I, for example, find the "one import per line" rule to be overly verbose.

My 0.02$ is that, no matter what we choose, we should be consistent in using one style and enforce it. The current formatting style wins in my opinion because rustfmt in the stable toolchain supports it so 1) it is easy to enforce (compared to nightly, as @Lorak-mmk pointed out) and 2) most developers will be familiar with this style.

If you want to improve readability, I think it is better to concentrate on other things that are not as easy to resolve mechanically: code structure, naming, documentation, etc.

Lorak-mmk:
> The current formatting style wins in my opinion because rustfmt in the stable toolchain supports it 

Afaik rustfmt stable isn't able to enforce any one style.

> Readability of the imports is a rather subjective thing. I, for example, find the "one import per line" rule to be overly verbose.

I don't find either style much greather than others (any I myself don't much care for readability of imports)).
One argument that convinces me for one-import-per-line is solving merge conflicts: there should be much less of them with this style, and import conflicts are vast majority of merge conflicts in Rust in my experience.

wprzytula (approved):
The argument of reducing number of merge conflicts convinced me.

PR #167, opened 2026-08-06
feat: make checkpoint table creation optional

### LINKED ISSUE #166: `TableBackedCheckpointSaver` requires `CREATE` permission even when the checkpoint table already exists
`TableBackedCheckpointSaver` constructor unconditionally run `CREATE TABLE IF NOT EXISTS` on the checkpoint table, and there is no way to skip it.

That breaks deployments where schema is managed out of band: the application's database user does not have `CREATE` permission on the `keyspace`. `IF NOT EXISTS` does not help the statement is still unauthorized DDL, so the saver cannot be constructed at all, even though every query it actually runs afterwards would succeed.

yacir:
Makes the checkpoint table creation in `TableBackedCheckpointSaver` optional, so the library can be used with a database user that has no `CREATE` permission and a table provisioned out of band.

Fixes #166 

Adding an `allow_table_creation: bool` to both constructors would break every existing caller. Instead this adds `TableBackedCheckpointSaverBuilder`.

`new` and `new_with_default_ttl` keep their exact signatures and are reimplemented on top of the builder. `create_table` defaults to `true`, so their behavior is unchanged. The builder mirrors `CDCLogReaderBuilder`.

### DIFF
--- a/scylla-cdc/src/checkpoints.rs
+++ b/scylla-cdc/src/checkpoints.rs
@@ -5,6 +5,7 @@ use crate::cdc_types::StreamID;
 use crate::cdc_types::Timestamp;
 use crate::cdc_types::make_idempotent_statement;
 use anyhow;
+use anyhow::Context;
 use async_trait::async_trait;
 use futures::FutureExt;
 use futures::future::RemoteHandle;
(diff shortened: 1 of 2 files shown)


scylla-cdc/src/checkpoints.rs
```diff
@@ -143,6 +154,134 @@ impl TableBackedCheckpointSaver {
...
+                "UPDATE {checkpoint_table} USING TTL {ttl}
+                SET generation = ?, time = ?
+                WHERE stream_id = ?"
+            ))
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to prepare the checkpoint statement, \
+                    make sure that the table {checkpoint_table} exists \
+                    and has the schema returned by TableBackedCheckpointSaver::table_schema_cql"
```
QuerthDP: Maybe the schema could've been printed here instead of the function reference?

QuerthDP (approved):
Looks really good! Thank you!

@wprzytula could you also take a look?

QuerthDP:
Additionally: please explain the motivation / changes done in the commit message not only in PR cover letter.

scylla-cdc/src/checkpoints.rs
yacir: You're right! That's much better. I've made the change.

wprzytula (approved):

QuerthDP:
Thanks for your contribution! 🙌

yacir:
@QuerthDP You're welcome! Any upcoming release planned?

QuerthDP:
> @QuerthDP You're welcome! Any upcoming release planned?

Yes. Possibly in 1-2 weeks. We're waiting for Scylla Rust Driver release.

QuerthDP:
@yacir [released](https://github.com/scylladb/scylla-cdc-rust/releases/tag/v0.7.0)

PR #168, opened 2026-08-10
treewide: fix dependency lower bounds and manage them from the workspace root

QuerthDP:
It was pointed out that `scylla-cdc` imports `tokio::task::JoinSet` while declaring `tokio = "1.1.0"`. `JoinSet` was not stabilised until 1.21, so the manifest advertised support for versions that cannot compile the crate. Auditing the rest of the manifests the same way turned up five more bounds below what the crates are actually built against.

| dep | was | now |
|---|---|---|
| tokio | 1.1.0 | 1.40 |
| chrono | 0.4.19 | 0.4.32 |
| tracing | 0.1.31 / 0.1.34 | 0.1.36 |
| async-trait | 0.1.51 / 0.1.52 | 0.1.56 |
| thiserror | 2.0 | 2.0.6 |
| anyhow | 1.0.48 | 1.0.51 |

The numbers come from `cargo +nightly update -Z direct-minimal-versions`, which pins every direct dependency to its declared minimum; each bound the resolver rejected was raised until the workspace resolved and built. All but `anyhow` land exactly on what `scylla` 1.3.1 requires — it is a non-optional dependency of all four crates, so its requirements are what really constrain the graph.

Worth flagging for tokio specifically: the fix is **1.40, not the 1.21** our own code needs. Because `scylla` 1.3.1 requires `tokio ^1.40`, no lockfile can resolve anything lower, so the reported compile failure is not reachable today. The declared bound was still wrong, and leaving a transitive dependency to enforce it is fragile.

The second commit then moves the versions into `[workspace.dependencies]`, since several were repeated across manifests and had to be kept in step by hand.

Fixes: VECTOR-776

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,31 @@ members = [
     "scylla-cdc-replicator",
     "scylla-cdc-test-utils",
 ]
+
+[workspace.dependencies]
+scylla-cdc = { version = "0.6.3", path = "scylla-cdc" }
+scylla-cdc-test-utils = { path = "scylla-cdc-test-utils" }
+
+anyhow = "1.0.51"
+async-trait = "0.1.56"
+chrono = "0.4.32"
+clap = { version = "4.5.9", features = ["derive"] }
+futures = "0.3.17"
+futures-util = "0.3.19"
+hex = "0.4.3"
+itertools = "0.13.0"
+num_enum = "0.7.2"
+rand = "0.8.5"
+rstest = "0.26.1"
+scylla = "1.3.1"
+thiserror = "2.0.6"
+tokio = { version = "1.40", features = [
+    "rt",
+    "io-util",
+    "net",
+    "time",
+    "macros",
+    "sync",
+] }
+tracing = "0.1.36"
+uuid = { version = "1.0.0", features = ["v1"] }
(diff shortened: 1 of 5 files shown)


wprzytula (commented):
❓ 
> The new bounds come from `cargo +nightly update -Z direct-minimal-versions`
verified with a build, and match what `scylla` 1.3.1 itself requires.

Is this part:
> and match what `scylla` 1.3.1 itself requires.

needed? Isn't it implied by the `cargo +nightly update -Z direct-minimal-versions`?

wprzytula (commented):
> The lockfile is unaffected.

It's not even versioned. Mentioning it is worthless.

QuerthDP:
> ❓
> 
> > The new bounds come from `cargo +nightly update -Z direct-minimal-versions`
> > verified with a build, and match what `scylla` 1.3.1 itself requires.
> 
> Is this part:
> 
> > and match what `scylla` 1.3.1 itself requires.
> 
> needed? Isn't it implied by the `cargo +nightly update -Z direct-minimal-versions`?

It is. It is not needed, I added it to mention that the higher tokio requirement than 1.21 come from `scylla` crate itself.

QuerthDP:
Changelog:
- adjusted commit messages

QuerthDP:
@wprzytula could you re-review?

wprzytula (approved):

